PR #21, opened 2024-02-29
feat(auth): Add single user OpenID Connect token provider APIs to support implicit flow 

samiralavi:
This PR adds the minimum set of APIs for the implicit flow of OpenID Connect to let the clients retrieve the access token and the ID token upon providing the right username and password. The username and password are defined in the settings module that reads them from the environment as variables. Unit tests are added to cover the APIs. Furthermore, some security packages are updated and obsolete user models are removed from the database layer. 

Note: The database schema is changed in a backward-compatible way, however, the `User` table has to be removed. I don't think a database migration script is needed at this time as other changes might come in future before the first stable release.


aqueductcore/backend/routers/auth.py
```diff
@@ -0,0 +1,151 @@
...
+    access_token = create_token(data={"sub": username}, expires_delta=access_token_expires)
+    id_token = create_token(data={"sub": username}, expires_delta=access_token_expires)
+
+    redirect_url = (
+        f"{redirect_uri}#access_token={access_token}"
+        f"&state={state}&id_token={id_token}&token_type=Bearer"
+    )
+
+    return RedirectResponse(
+        url=redirect_url,
```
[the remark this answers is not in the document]
samiralavi: This is intentional until the redirect URI from the frontend is finalised. Once this happens, a validation step will be added.

aqueductcore/backend/routers/auth.py
```diff
@@ -0,0 +1,152 @@
...
+        username: str = str(payload.get("sub"))
+        if username is None:
+            raise credentials_exception
+    except JWTError as exc:
+        raise credentials_exception from exc
+    user = User(username=username)
+    return user
+
+
+@router.get("/.well-known/openid-configuration")
```
jatinriverlane, line 97: @samiralavi In these URLs, when I try to send a request, I'm getting error related to `StaticFiles directory 'frontend_build' does not exist.`

Is there some way to test it ? Also, is there some more configuration needed to run single user auth ?

safeamiiir (approved):
From my point of view, it looks legit but I'll ask for changes when single user is provided in the frontend.

aqueductcore/backend/routers/auth.py
samiralavi: @jatinriverlane the `frontend_build` directory gets generated as a result of `build_frontend` scripts that build the react files and generate static files to be served by the server.

samiralavi: Don't forget to use the `/api/` prefix with the URL and the HTTP verb you are using

samiralavi:
@safeamiiir sure, we will revisit this during integration with the frontend. @jatinriverlane please let me know if you are happy with the changes to proceed with the merge. Thanks.

jatinriverlane (approved):
Looks fine Amir, we can implement any further change required during integration with frontend.

PR #22, opened 2024-03-05
fix(experiment)!: Change experiment generated alias format to `YYYYMMDD-[counter that resets per day]`

samiralavi:
This PR changes the default behaviour for assigning the alias of the experiment when a new experiment is created. The new alias format is `YYYYMMDD-[counter that resets per day]`. My solution is more on the robustness side rather than performance as we need a way to know what was the last counter value used. As the backend should be scalable, this information should be distributed across the instances and the API should be stateless. Therefore, every time an experiment is to be created, we first query the experiments (only alias column) in the current date (API server time) and then sort them in the software to find the last used counter index. Then it is increased by one and the experiment is created. In my opinion, this will be ok for the foreseeable future.

The other change in the PR is refactoring the service layer with Pydantic validations instead of manual validation functions that required proper testing for corner cases. Some files were also formatted incorrectly in previous PRs, and there are fixed on the files that I need to change the code.

Note: I have safely implemented the method so that old data with a different alias pattern will not cause any problems, so the database doesn't require migration.


aqueductcore/backend/routers/auth.py
```diff
@@ -0,0 +1,152 @@
...
+    access_token = create_token(data={"sub": username}, expires_delta=access_token_expires)
+    id_token = create_token(data={"sub": username}, expires_delta=access_token_expires)
+
+    redirect_url = (
+        f"{redirect_uri}#access_token={access_token}"
+        f"&state={state}&id_token={id_token}&token_type=Bearer"
+    )
+
+    return RedirectResponse(
+        url=redirect_url,
```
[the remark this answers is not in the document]
samiralavi: This is intentional until we fix the redirect URL to a known address.

aqueductcore/backend/services/experiment.py
```diff
@@ -107,6 +116,7 @@ async def get_all_experiments(  # pylint: disable=too-many-arguments
     return [(await experiment_orm_to_model(item)) for item in result.unique().scalars().all()]
 
 
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
```
gabrielg42, line 119: What does this do?

aqueductcore/backend/services/utils.py
```diff
@@ -53,11 +53,11 @@ def tag_model_to_orm(value: TagCreate) -> orm.Tag:
     return tag
 
 
-def generate_id_and_alias() -> Tuple[UUID, str]:
+def generate_id_and_alias(experiment_index: int) -> Tuple[UUID, str]:
     """Generate uuid and alias for experiment"""
     current_date = datetime.today().strftime("%Y%m%d")
     experiment_id = uuid4()
-    alias = f"{current_date}-{str(experiment_id)[-12:]}"
+    alias = f"{current_date}-{experiment_index}"
```
gabrielg42, line 60: Does the sorting still work even if you have indices with different numbers of digits. E.g. is `"2"` < `"11"`?

aqueductcore/backend/services/experiment.py
samiralavi: This is a Pydantic method of validating arguments passed to the function it decorates. By default, Pydantic expects all arguments of the Pydantic model or primitive Python types, and if it faces a different type it complains. `arbitrary_types_allowed` fixes this issue to not check types that are not of those, here it is the database session object.

aqueductcore/backend/services/utils.py
gabrielg42: Right, but what I mean to ask is, does the sorting work properly even though this index is stored as a string? For integers, `2 < 11`, but for strings `"2" > "11"`.

samiralavi: Sorry I don't understand properly, there is no sorting in the `generate_id_and_alias` function. If you mean sorting when we retrieve the experiments of the day, then yes, I already considered that. If you go to line 254 of  `aqueductcore/backend/services/experiment.py`, you can see that the string is split and the index is converted to an integer for the key of the sorting function. Sorting happens outside of `generate_id_and_alias`.

aqueductcore/backend/services/experiment.py
```diff
@@ -199,27 +212,52 @@ async def get_experiment_files(
...
+
+    if not today_experiments:
+        experiment_id, alias = generate_id_and_alias(experiment_index=1)
+    else:
+        last_experiment_alias = sorted(today_experiments, key=lambda x: int(x.split("-")[-1]))[-1]
+        if re.match(EXPERIMENT_ALIAS_PATTERN, last_experiment_alias):
+            index = int(last_experiment_alias.split("-")[-1])
+            experiment_id, alias = generate_id_and_alias(experiment_index=index + 1)
+        else:
+            experiment_id, alias = generate_id_and_alias(experiment_index=1)
```
jatinriverlane, line 259: I guess this can be changed to 

```python
    today = datetime.combine(date.today(), time.min)
    today_experiments_statement = (
        select(orm.Experiment.alias)
        .filter(orm.Experiment.created_at >= today)
        .where(orm.Experiment.alias.regexp_match(EXPERIMENT_ALIAS_PATTERN))
        .order_by(orm.Experiment.created_at.desc())
        .limit(1)
    )

    today_experiments = (await db_session.execute(today_experiments_statement)).scalars().all()
    print(today_experiments)

    if not today_experiments:
        experiment_id, alias = generate_id_and_alias(experiment_index=1)
    else:
        last_experiment_alias = sorted(today_experiments, key=lambda x: int(x.split("-")[-1]))[0]
        if re.match(EXPERIMENT_ALIAS_PATTERN, last_experiment_alias):
            index = int(last_experiment_alias.split("-")[-1])
            experiment_id, alias = generate_id_and_alias(experiment_index=index + 1)
        else:
            experiment_id, alias = generate_id_and_alias(experiment_index=1)
```

This way, we are not getting all the experiments for the day and then selecting the last last experiment by the `-1` index. Instead, we are running the query with limit 1 and sorting by `created_at` in descending order, hence getting last experiment of the day.

samiralavi: Re-reading my reply again, I found that even my solution is not going to address these challenges! Thanks @jatinriverlane for this. I will re-think it.

aqueductcore/backend/services/utils.py
samiralavi: I think the solution is to use an advisory lock! A simpler version instead of using Redis.
https://medium.com/inspiredbrilliance/a-practical-guide-to-using-advisory-locks-in-your-application-7f0e7908d7e9

aqueductcore/backend/services/experiment.py
jatinriverlane: Sure, thanks @samiralavi I'll also think about a solution for this

jatinriverlane (approved):
Thanks @samiralavi

aqueductcore/backend/services/experiment.py
jatinriverlane: A point to note here is that the labs will not be running such high number of experiments that there are multiple experiments running even every few seconds, here the accuracy is of milliseconds. This might cause issues in the systems with thousands if not millions of active users.

gabrielg42 (approved):

safeamiiir (approved):

PR #26, opened 2024-03-11
fix(frontend): Disable frontend static files caching as a prerequisite for authentication

samiralavi:
Disabling caching on the browser is required to make sure the authentication flow goes smoothly when the user is logged out. Due to the way react works by playing with the routes on the browser, if caching was not disabled, a cached version of the frontend would be returned with frontend parsing of the URL although the user was already logged out! This PR also removes the obsolete passlib library from the dependencies and uses bcrypt directly.


safeamiiir (approved):
Thanks! very helpful before frontend changes.

PR #27, opened 2024-03-13
feat(security)!: Permission based access to the API via authentication tokens

samiralavi:
This PR adds permissions processing retrieved from the tokens in the service layer. The dependency injection system of FastAPI is used to propagate the scopes of the users down to the service layer to implement the business logic of permission-based access for the single user. In the core version, the token is not validated, however, in the pro version, I have implemented the required functionality for the tokens and users.

### DIFF
--- a/aqueductcore/backend/context.py
+++ b/aqueductcore/backend/context.py
@@ -1,22 +1,75 @@
 """Server context management module."""
 
-from typing import AsyncGenerator
+from enum import Enum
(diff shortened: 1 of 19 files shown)


gabrielg42 (approved):

jatinriverlane (approved):
Thanks @samiralavi , everything seems fine and has covered all the cases.

PR #28, opened 2024-03-13
feat(delete_experiment): add feature to delete experiment

jatinriverlane:
# Delete Experiment Feature
Allows users to delete experiments along with their linked files

## ⚠️  Breaking Changes
There are no breaking changes in this PR

## ✔️ Checklist
- [x] **Code Review**: I have reviewed my changes to ensure they meet the project's coding standards and guidelines.
- [x] **Sanity Checks**:  I have performed basic sanity checks to verify the correctness of my changes.
- [x] **Testing**: Tests have been added or updated to cover the changes made in this pull request, and all tests passed successfully.
- [x] **Style Guidelines**: My code follows the style guidelines of this project

### DIFF
--- a/aqueductcore/backend/routers/graphql/mutations_schema.py
+++ b/aqueductcore/backend/routers/graphql/mutations_schema.py
@@ -20,7 +20,7 @@ from aqueductcore.backend.routers.graphql.mutations.experiment_mutations import 
     remove_tag_from_experiment,
     update_experiment,
 )
-from aqueductcore.backend.routers.graphql.types import ExperimentData, GenericResponse
+from aqueductcore.backend.routers.graphql.types import ExperimentData
 
 
 @strawberry.type
@@ -76,14 +76,12 @@ class Mutation:
         )
         return experiment
 
-
     @strawberry.mutation
     async def remove_experiment(
         self, info: Info, experiment_remove_input: ExperimentRemoveInput
-    ) -> GenericResponse:
+    ) -> None:
         """Mutation to remove experiment"""
         context = cast(ServerContext, info.context)
-        result, message = await remove_experiment(
+        await remove_experiment(
             db_session=context.db_session, experiment_remove_input=experiment_remove_input
         )
-        return GenericResponse(result=result, message=message)
--- a/aqueductcore/backend/routers/graphql/mutations_schema.py
+++ b/aqueductcore/backend/routers/graphql/mutations_schema.py
@@ -9,16 +9,18 @@ from strawberry.types import Info
 from aqueductcore.backend.context import ServerContext
 from aqueductcore.backend.routers.graphql.inputs import (
     ExperimentCreateInput,
+    ExperimentRemoveInput,
     ExperimentTagInput,
     ExperimentUpdateInput,
 )
 from aqueductcore.backend.routers.graphql.mutations.experiment_mutations import (
     add_tag_to_experiment,
     create_experiment,
+    remove_experiment,
     remove_tag_from_experiment,
     update_experiment,
 )
-from aqueductcore.backend.routers.graphql.types import ExperimentData
+from aqueductcore.backend.routers.graphql.types import ExperimentData, GenericResponse
 
 
 @strawberry.type
@@ -73,3 +75,15 @@ class Mutation:
             db_session=context.db_session, experiment_tag_input=experiment_tag_input
         )
         return experiment
+
+
+    @strawberry.mutation
+    async def remove_experiment(
+        self, info: Info, experiment_remove_input: ExperimentRemoveInput
+    ) -> GenericResponse:
+        """Mutation to remove experiment"""
+        context = cast(ServerContext, info.context)
+        result, message = await remove_experiment(
+            db_session=context.db_session, experiment_remove_input=experiment_remove_input
+        )
+        return GenericResponse(result=result, message=message)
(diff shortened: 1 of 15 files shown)


safeamiiir (approved):
We're removing files, experiment tags and experiment record.
That's all we need to do I guess.
Thanks Jatin!

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -73,3 +75,15 @@ async def remove_tag_from_experiment(
             db_session=context.db_session, experiment_tag_input=experiment_tag_input
         )
         return experiment
+
+
+    @strawberry.mutation
+    async def remove_experiment(
+        self, info: Info, experiment_remove_input: ExperimentRemoveInput
+    ) -> GenericResponse:
```
samiralavi, line 83: The return type here is redundant and it can be `None`. In a scenario where an exception is raised in this function, graphql returns the error message in the `error["details"]` field. The return will succeed only when the experiment removal is done correctly (for example after checking permissions) and in that case, there is no need to return anything to the user. Please set the return type to `None` and remove the `GenericResponse` type totally to simplify the process.

aqueductcore/backend/routers/graphql/types.py
```diff
@@ -77,3 +77,11 @@ class Tags:
 
     tags_data: List[str] = strawberry.field(description="The list of tags.")
     total_tags_count: int = strawberry.field(description="Total number of tags.")
+
+
+@strawberry.type(description="Generic response type")
+class GenericResponse:
```
samiralavi, line 83: This is not needed.

aqueductcore/backend/services/experiment.py
```diff
@@ -413,3 +414,43 @@ async def create_tag(db_session: AsyncSession, tag: TagCreate) -> TagRead:
...
+    if not get_experiment_result.scalars().first():
+        raise AQDDBExperimentNonExisting("DB query failed due to non-existing experiment with " + \
+                                         "the specified ID.")
+
+    folder_path = build_experiment_dir_absolute_path(
+        experiments_root_dir=str(settings.experiments_dir_path), experiment_id=experiment_id
+    )
+
+    try:
+        if os.path.exists(folder_path):
```
samiralavi, line 434: This is slow and not safe as multiple steps are happening, which should be atomic. Please use `[shutil.rmtree()](http://docs.python.org/library/shutil.html#shutil.rmtree)` to remove the experiment directory in a single call (although it uses `os.remove` internally but it is optimised by the community). You might need to check if the directory exists, but a simpler approach is to pass `ignore_errors=True` so if the directory doesn't exist, it won't fail.

tests/unittests/test_graphql_mutations.py
```diff
@@ -386,3 +423,26 @@ async def test_remove_tag_from_experiment(
...
+        db_session.add(db_exerperiment)
+        await db_session.commit()
+        await db_session.refresh(db_exerperiment)
+
+    schema = Schema(query=Query, mutation=Mutation)
+
+    context = ServerContext(db_session=db_session)
+    resp = await schema.execute(remove_experiment_mutation, context_value=context)
+
+    assert resp.errors is None
```
samiralavi, line 447: The assertions are not enough, you need to check if the files are actually getting removed along side the experiment in the database.

samiralavi (changes requested):
Thanks @jatinriverlane for this work. I have requested minor but important changes. Also, the tests don't have enough assertions to check if the experiment is removed properly.

aqueductcore/backend/routers/graphql/mutations_schema.py
jatinriverlane: @samiralavi We would need to have some type in response as strawberry expects some strawberry field in response. On setting None as response type, it's throwing this error as well
```Field 'removeExperiment' must not have a selection since type 'Void' has no subfields.```

File: aqueductcore/backend/services/experiment.py (hunk 3 above)
Comment (jatinriverlane) [reply aqueductcore/backend/services/experiment.py:434]: @samiralavi This would be a very good and fast way, I've updated it to delete folder and it's files using shutil.rmtree

File: tests/unittests/test_experiment_services.py [hunk 5]
```diff
@@ -437,3 +439,39 @@ async def test_get_experiment_files_empty(
...
+    db_session: AsyncSession, experiments_data: List[ExperimentCreate]
+):
+    for experiment in experiments_data:
+        db_experiment = experiment_model_to_orm(experiment)
+        db_session.add(db_experiment)
+
+    await db_session.commit()
+
+    experiment_id = uuid4()
+    with pytest.raises(AQDDBExperimentNonExisting):
```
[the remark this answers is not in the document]
jatinriverlane: Have updated tests to confirm deletion of folder as well in this test, thanks Amir

tests/unittests/test_graphql_mutations.py
jatinriverlane: Have updated tests to confirm deletion of folder as well in this test, thanks Amir

aqueductcore/backend/routers/graphql/mutations_schema.py
samiralavi: @jatinriverlane I now understand the point. Can you instead of `GenericResponse`, just return the ID of the experiment deleted? The current `GenericResponse` is redundant the basic contract is if the mutation doesn't fail, then it has succeeded. Duplicating the `result` in the response is redundant and might confuse API users that there is a chance that a mutation has succeeded and failed at the same time!

samiralavi (approved):
Thanks @jatinriverlane for the change. I want to still insist on removing the generic response. Please see my comment and check if you see my point valid

PR #33, opened 2024-03-27
feat(user): Add `UserInfo` node to the GraphQL endpoint to return username and scopes.

samiralavi:
This PR adds a new node to the GraphQL endpoint that returns the username and scopes from the provided token. The following snippet shows the available query.
```
query TestQuery {
  getCurrentUserInfo {
    scopes
    username
  }
}
```
`username` is a string type, and `scopes` is a list of string. The following screenshot shows the GraphiQL execution:

image

### DIFF
--- a/aqueductcore/backend/context.py
+++ b/aqueductcore/backend/context.py
@@ -4,13 +4,12 @@ from enum import Enum
 from typing import AsyncGenerator, Set
 from uuid import UUID
 
+from aqueductcore.backend.session import get_session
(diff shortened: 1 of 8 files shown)


safeamiiir (approved):
Let's just consider changing the hard-coded username into a dynamic one later.
Other than that I'm happy with it, thanks Amir!

jatinriverlane (approved):
Thanks @samiralavi

PR #36, opened 2024-04-09
fix(database): Fix database session getting unusable after first commit.

samiralavi:
This PR fixes the issue with the async database session that gets closed after the first commit and stops the automatic beginning of new transactions in consequent calls to the session. In this PR, the database migration script with the new schema is added and the README.md file for the database migration guide is updated.


samiralavi:
I don't know why eslint fails, not related to changes in this PR.

jatinriverlane (approved):
Tested it, it's working fine. Great that migrations are handled by Alembic now. Thanks @samiralavi

PR #39, opened 2024-04-15
feat(setting): Add a new optional environment variable for Keycloak URL

samiralavi:


aqueductcore/backend/settings.py
```diff
@@ -41,20 +41,11 @@ class Settings(BaseSettings):
...
-
     default_username: str = "admin"
     """Default username for the User"""
 
-    token_secret: str = secrets.token_hex(32)
-    """Tokens secret."""
-
-    tokens_expiry_time_minutes: int = 30
-    """Authentication token expiry time in minutes."""
+    keycloak_url: Optional[HttpUrl] = None
```
jatinriverlane, line 47: Should we use name containing `keycloak` ? I guess this can be mentioned as `auth_url`...

jatinriverlane (approved):
Thanks @samiralavi

aqueductcore/backend/settings.py
samiralavi: We decided to use `keycloak` in the name.

Issue #40, opened 2024-04-15
New feature: being able to execute plugins


str-anger:
I want Aqueduct being able to execute plugin-like code. This includes:
- having infrastructure (GraphQL endpoints) to store and list plugins
- having code to execute a particular plugin function (also, GraphQL endpoint).
- having simple examples to implement my own plugins.

samiralavi:
CTRL-2732

PR #41, opened 2024-04-15
feat(plugins): adding plugin functionality to the server app, issue #40

str-anger:
In this PR am adding base classes looking in both ways:
1. They enable aqueduct to list and execute plugins using GraphQL
2. They allow to run plugins in the post simple way


samiralavi (commented):
Thanks @str-anger for this PR. I see it as a robust plugin execution solution. After all, in this model, plugins are scripts executed by the OS as a subprocess.

I have some doubts about the use of PyAqueduct for the plugins. PyAqueduct is a user-facing Python library to interacts with the "User" API of Aqueduct. I don't know to how much extent, plugins need more privileged access to the internal functionality of Aquecut. There are a few other points here, in addition to my inline comments:

- You have created the PR against the main branch. Our main branch should be in a releasable state, but this PR is a work in progress.
- For future branches please follow the branch naming guideline that @safeamiiir has created. 
- About the first plugin there, the markdown converter, I like to know the process that led to this design. We have not yet captured the requirements for this feature, and there are other ways to achieve it. As there is no product requirement on this, then let's not have it in this PR.

samiralavi:
> Didn't get - should I move this to Draft or should I branch from some different branch?
Yes please create a new branch for example `feat/plugins` and merge the PRs there then once the feature is ready, we merge it into the main. The sooner we merge this branch into the main, the better.

samiralavi (approved):
Thanks @str-anger, all is good now. Great work.

PR #52, opened 2024-04-18
feat(security): add scopes logic in the frontend

safeamiiir:
Add scopes in the frontend


samiralavi (approved):
Great work, thanks @safeamiiir.

PR #54, opened 2024-04-18
feat(tags): Added support for more special characters in tags

jatinriverlane:
# Changes
* Support for more special characters in tags namely ":" and "/".
* Separate error messages for over length tags and unsupported special characters.


samiralavi (approved):

safeamiiir (approved):
Thanks Jatin! it's also worth modifying and adding some small tests for these special characters.

PR #57, opened 2024-04-22
fix(timezone): Add timezone awareness to time operations based on UTC.

samiralavi:
PostreSQL datetime type doesn't include the time zone information and the recommended approach is that the application should know the time zone when it stores the data and retrieves it. Aqueduct will now keep the time zone as UTC when stored in the database. Furthermore, the API will be based on UTC time zones. In the API returned values, the serialized datetime comes with time zone information included (i.e. for UTC `+00:00`). The input types such as the date filter input type accepts serialized datetime in any time zone (ISO format); it gets updated to  UTC time zone before being used by the service layer. This PR adds the required tests to consider start and stop datetime.


pyproject.toml
```diff
@@ -52,6 +52,7 @@ alembic = "^1.12"
 psycopg2-binary = "^2.9"
 mkdocstrings = {version = "^0.24",extras = ["python"]}
 nbconvert = "^7.14"
+pytz = "^2024.1"
```
str-anger, line 55: I'd disapprove this change. You are bringing a dependency only for the sake of the single test which creates a time zone by the name. I think you might avoid adding a new package, OR, at least make it optional and add to extras, such that the non-dev version won't require adding this library.

samiralavi: I understand your point about dependencies and I agree with that, but this is a development dependency in the `toml` file defined under `[tool.poetry.group.dev.dependencies]`. Could you review this again?

aqueductcore/backend/services/experiment.py
```diff
@@ -90,15 +90,15 @@ async def get_all_experiments(  # pylint: disable=too-many-arguments
         )
 
     if start_date is not None and end_date is not None:
-        start_date = datetime.combine(start_date, time.min)
```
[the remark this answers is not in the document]
samiralavi: old logic 🗡️ 😆

str-anger (approved):

aqueductcore/backend/services/experiment.py
```diff
@@ -90,15 +90,15 @@ async def get_all_experiments(  # pylint: disable=too-many-arguments
...
-        end_date = datetime.combine(end_date, time.max)
-        statement = statement.filter(orm.Experiment.created_at.between(start_date, end_date))
+        statement = statement.filter(
+            orm.Experiment.created_at.between(start_date.astimezone(timezone.utc), end_date)
```
safeamiiir, line 94: why are you not also using `end_date.astimezone(timezone.utc)`?

safeamiiir (approved):

aqueductcore/backend/services/experiment.py
samiralavi: Good catch, thanks, it was missing here! It seems the tests I have written did not cover this bug. Will fix.

jatinriverlane (approved):

PR #58, opened 2024-04-22
fix(time): fix date and time to a local version

safeamiiir:
Fix date and time, it's now showing local time


jatinriverlane (approved):
This is working as expected, thanks @safeamiiir

samiralavi (approved):
Thanks @safeamiiir

Issue #63, opened 2024-04-22
Implement dummy/example python plugin


str-anger:
Decide with @tomhartley  and @samiralavi what it should look like.

PR #71, opened 2024-04-24
#63 adding a python example

str-anger:
This PR adds a simple plugin which demonstrates a few plugin capabilities:
1. ability to install custom dependencies for plugin execution
2. access to remote service protected with the key
3. access to aqueduct using `pyaqueduct` library.
4. using common presets for all functions of the plugin


tests/unittests/test_plugin_executor.py
```diff
@@ -195,9 +204,25 @@ def test_plugin_validation_ok(self, plugin):
...
+                params={},
             ),
         ],
     )
     def test_plugin_validation_raises(self, plugin):
         with pytest.raises(AQDValidationError):
             PluginExecutor._validate_plugin(plugin)
+
+    @pytest.mark.skip
+    def test_plugin_wolfram_alpha(self):
```
samiralavi, line 216: Not in this PR but later on, the plugins tests should be a separate file.

plugins/python-example/manifest.yml
```diff
@@ -0,0 +1,55 @@
...
+    # https://stackoverflow.com/a/54482177
+    #
+    # This thick allows to avoid `aqueductcore`
+    # dependency on client library.
+    #
+    # Plugin may cache some data and librarues
+    # for it's further usage, or, alternatively,
+    # it may drop and recreate
+    # its environment on each run.
+    script: >
```
samiralavi, line 35: This can be problematic. Aqueduct is a web application that is shipped via a docker image. What is your proposal for shipping plugins to the users? How this custom Python environment can be used in a container?

samiralavi (approved):
Thanks @str-anger this looks good. I approve with the condition that you provide a good proposal for the delivery of plugins to the users in the reply to my comment.

plugins/python-example/manifest.yml
str-anger: @samiralavi what stops user from running pip install inside a container? I am new to containers, but I feel like if they are allowed to create files, they should be allowed to create venvs. Right?

samiralavi: That is the ugliest way. Just ask our DevOps engineers. We shouldn't expect anyone to enter the docker container for a normal configuration. Everything should be configurable before bringing up the container. A possible approach is mounting a volume for the plugins into the container, so the user can change the plugin files. They have to restart the container for the changes to take effect as plugins will be loaded during startup.

The main issue/challenge is custom Python environments. You can't inject a Python environment into a container via volumes. It might work but it is a hack and requires some prerequisites like the same Python version, host OS dependencies, and CPU architecture. Also again, it is not scalable.

str-anger: > That is the ugliest way

This is not beautiful, no one ever promised this to be beautiful with short deadlines. Job scheduling and isolation is "nice to have" and planned for later. Plugins need to be able to customise their environments to be able to perform functions. If we (in the default "must have" behaviour) agree on process spawning, **please propose another way to customise execution environment**. I will implement it as a part of this PR. Please keep in mind that pyaqueduct and aqueduct core from time to time are incompatible in dependencies.

> We shouldn't expect anyone to enter the docker container for a normal configuration

I don't see how someone "enters" the container. Probably you are talking about mounting the folder with plugins externally as a volume? Here is no contradiction to current implementation.

> They have to restart the container for the changes to take effect as plugins will be loaded during startup.

I can't see any problem. Also, are volumes immutable? I think user puts files into folder, and these files are now visible in a folder inside a container.

> You can't inject a Python environment into a container via volumes

I never proposed this.

Currently I feel like you dislike the original design, so I want to have a proposal, or we keep it for now. I myself want to switch to a job execution, e.g. https://docs.celeryq.dev/en/stable/, but I feel a little bit sqeezed between different requirements coming from different sides.

PR #73, opened 2024-04-24
feat(table): make file table sortable

safeamiiir:
to make the list of files in the experimentDetails page sortable.

### DIFF
--- a/aqueductcore/frontend/src/helper/functions.tsx
+++ b/aqueductcore/frontend/src/helper/functions.tsx
@@ -2,6 +2,7 @@ import { FocusEvent } from "react";
 import dayjs from "dayjs";
 
 import { SortOrder } from "types/componentTypes";
+import { ExperimentFileType } from "types/globalTypes";
 
 // ################## DOM related functions ################## //
 export const focusInCurrentTarget = ({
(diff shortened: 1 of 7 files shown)


aqueductcore/frontend/src/helper/functions.tsx
```diff
@@ -15,3 +19,53 @@ export const focusInCurrentTarget = ({
...
+  if (itemB < itemA) {
+    return -1;
+  }
+  if (itemB > itemA) {
+    return 1;
+  }
+  return 0;
+}
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
samiralavi, line 44: Why linting is disabled on these functions?

samiralavi (approved):
Thanks @safeamiiir. The PR looks, good. I assume there is a default sorting, either by name or by date when the experiment page is loaded.

aqueductcore/frontend/src/helper/functions.tsx
safeamiiir: fixed

safeamiiir:
> Thanks @safeamiiir. The PR looks, good. I assume there is a default sorting, either by name or by date when the experiment page is loaded.

yes, this is the default behaviour that the most recently modified one will be more on top.

jatinriverlane (approved):
Thanks @safeamiiir

Issue #77, opened 2024-04-25
Implement changes to plugin manifest processing and interfaces


str-anger:
1. Support default values in variable definition
2. Support bool type (checkbox)
3. multiline -> textarea
4. Dropdown strings (select -> option?)
5. Save log into an experiment; return a file name
6. Expose a single experiment for a function

PR #79, opened 2024-04-26
Task #77: plugins changes

str-anger:
Support default values in variable definition
Support bool type (checkbox)
 multiline -> textarea
Dropdown strings (select -> option?)
Save log into an experiment; return a file name
Expose a single experiment for a function


aqueductcore/backend/routers/graphql/types.py
```diff
@@ -119,6 +122,8 @@ class PluginExecutionResult:
     return_code: int
     stdout: str
     stderr: str
+    log_exeperiment: str
```
gabrielg42, line 125:
```suggestion
    log_experiment: str
```

plugins/python-example/manifest.yml
```diff
@@ -41,15 +41,18 @@ functions:
...
-        data_type: str
-      - !parameter
-        name: experiment
-        description: experiment to store result
-        data_type: experiment
-      - !parameter
-        name: result_file
-        description: file name to store result
-        data_type: str
+      - !!python/object/apply:aqueductcore.PluginParameter
```
gabrielg42, line 44: Is there a cleaner way for the user to write this?

aqueductcore/backend/models/plugin.py
```diff
@@ -28,20 +48,95 @@ class PluginParameter(yaml.YAMLObject):
     """Typed and named parameter of the plugin function"""
 
     yaml_tag = "!parameter"
-    yaml_loader = yaml.SafeLoader
+    yaml_loader = yaml.UnsafeLoader
```
safeamiiir, line 51: I don't see where these two variables might be used? Are they only to load?
also same question for them in "function" and "parameter".

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -89,13 +97,15 @@ async def remove_experiment(
     @strawberry.mutation
     async def execute_plugin(
             self,
+            info: Info,
             plugin: str,
             function: str,
             params: List[List[str]],
     ) -> PluginExecutionResult:
```
safeamiiir, line 104: A broader question than this line, but what about assigning a `plugin-id` to a plugin?
They could give it whatever name they want, but something like `DummyPlugin` instead of `Dummy plugin` reduces future issues and also lets the plugin names be duplicated.

safeamiiir (commented):
I was just wondering if we can/need to have a `pluginId` instead of directly using their names

aqueductcore/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""`aqueductcore` is a module with backend and frontend implementation for Aqueduct software."""
+
+from aqueductcore.backend.models.plugin import PluginParameter
```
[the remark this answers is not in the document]
str-anger: It was for parsing reason. This will go away after I will use @samiralavi's method.

aqueductcore/backend/plugins/plugin_executor.py
```diff
@@ -5,104 +5,20 @@
...
     Plugin,
     PluginExecutionResult,
-    PluginFunction,
-    PluginParameter,
 )
-from aqueductcore.backend.errors import AQDValidationError
 from aqueductcore.backend.settings import settings
 
 
-class SupportedTypes(Enum):
```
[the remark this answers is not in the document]
str-anger: Initially I was relying on the simple-to-no validation, and Plugin, etc I considered as data classes. But later I realised that more and more logic is required from them, so they become important entities. Entities incapsulate their behaviour according to OOP principles =)

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -105,9 +115,39 @@ async def execute_plugin(
...
+        if exp_parameter is None:
+            raise AQDValidationError(f"Function {plugin}/{function} has no experiment parameters")
+        exp_id = dict_params[exp_parameter.name]
+        result = function_object.execute(plugin_object, dict_params, timeout=600)
+        experiment = await get_experiment_by_alias(
+            user_info=context.user_info,
+            db_session=context.db_session,
+            alias=exp_id
+        )
+        now = datetime.datetime.now(tz=datetime.timezone.utc).strftime("%Y%m%d-%H%M%S")
```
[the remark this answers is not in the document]
str-anger: You means just `now()`?

plugins/python-example/manifest.yml
str-anger: Yes, Amir showed me, will implement soon

aqueductcore/backend/models/plugin.py
str-anger: This will all change after I switch to pydantic =)
These field are used implicitly by yaml parser, they define(d) a strategy of parsing (Safe/Unsafe) and a tag used inside manifest.

aqueductcore/backend/routers/graphql/mutations_schema.py
str-anger: Let us discuss this. Plugins are now something without versioning and naming. I will draw attention of @gabrielg42 to this, but honestly I would not restrict users too much on this.

safeamiiir (approved):

aqueductcore/backend/models/plugin.py
```diff
@@ -24,36 +44,90 @@ class PluginExecutionResult(BaseModel):
...
+    options: Optional[List[str]] = None
+
+    def __str__(self):
+        return f"{self.display_name} ({self.name}: {self.data_type})"
+
+    def __repr__(self):
+        return f"<{str(self)}>"
+
+    # pylint: disable=too-many-return-statements,too-many-branches
+    def validate_object(self):
```
samiralavi, line 64: You don't need this method as a custom validator for fields as long as they have no dependency. Pydantic automatically checks if the field is optional or has to be provided and complains accordingly.

aqueductcore/backend/models/plugin.py
```diff
@@ -24,36 +44,90 @@ class PluginExecutionResult(BaseModel):
...
+                f"but was {self.data_type}"
+        )
+        if self.default_value:
+            self.default_value = self.validate_value(self.default_value)
 
-    def __init__(self, name: str, description: str, data_type: str):
-        self.name = name
-        self.description = description
-        self.data_type = data_type
+    def validate_value(self, value: str) -> str:
```
samiralavi, line 78: You can just assign the type of `data_type` member to `SupportedTypes`, after making sure it is a string enum and Pydantic saves your time here too 🙈

aqueductcore/backend/models/plugin.py
```diff
@@ -72,6 +146,7 @@ def execute(
             PluginExecutionResult: OS process results.
         """
         my_env = os.environ.copy()
+        self.validate_values(params)
```
samiralavi, line 149: Pydantic object instantiation is the actual validation, so no extra validation call is needed.

aqueductcore/backend/models/plugin.py
```diff
@@ -131,8 +231,31 @@ def from_folder(cls, path: Path) -> Plugin:
...
+
+        Returns:
+            Plugin function object
+        """
+        functions = [f for f in self.functions if f.name == name]
+        if len(functions) != 1:
+            raise AQDValidationError(f"There should be exactly 1 function with name {name}")
+        return functions[0]
+
+    def validate_object(self):
```
samiralavi, line 256: Here too, with proper choice of type hints, Pydantic does it all for you.

aqueductcore/backend/models/plugin.py
```diff
@@ -3,19 +3,39 @@
...
+    FLOAT = "float"
+    EXPERIMENT = "experiment"
+    FILE = "file"
+    BOOL = "bool"
+    SELECT = "select"
+
+    @staticmethod
+    def values() -> Set[str]:
+        """String values of the enum members."""
+        return set(map(str, SupportedTypes._value2member_map_))
```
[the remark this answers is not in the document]
samiralavi: It helps you later when using type hints for Pydantic!

aqueductcore/backend/models/plugin.py
```diff
@@ -96,27 +171,52 @@ def execute(
...
             )
 
+    def get_default_experiment_parameter(self) -> Optional[PluginParameter]:
+        """Return first experiment variable defined in manifest.
+
+        Return:
+            plugin parameter object.
+        """
+        for variable in self.parameters:
+            if variable.data_type == SupportedTypes.EXPERIMENT.value:
```
samiralavi, line 181: I guess this can also be enforced by type hint too.

aqueductcore/backend/models/plugin.py
```diff
@@ -96,27 +171,52 @@ def execute(
...
+        return None
+
+    def validate_object(self):
+        """Validate the instance of the function and its parameters."""
+        if not self.name:
+            raise AQDValidationError("Plugin function should have a name.")
+        for param in self.parameters:
+            param.validate_object()
+
+    def validate_values(self, params: Dict[str, str]):
```
samiralavi, line 192: Same here, give it a try

tests/unittests/test_plugins.py
```diff
@@ -0,0 +1,113 @@
...
+                                data_type="something",
+                                default_value="some",
+                            )
+                        ],
+                    ),
+                ],
+            ),
+        ],
+    )
+    def test_plugin_validation_raises(self, plugin):
```
samiralavi, line 105: If you use pydantic validation, you can omit these tests. The more we rely on Pydantic, it reduces our test surface and chances of bugs in validations.

samiralavi (approved):
Thanks @str-anger, it is great work. I am happy with changes, just please take a look at the minor comments I had about better use of Pyndatic.

aqueductcore/backend/models/plugin.py
str-anger: I still need some validations on data coming from the UI, but we may discuss design in person

PR #83, opened 2024-04-29
fix(readme): Add links for `aqueductcore-dev` images on Dockerhub to README.md

samiralavi:
This PR adds the information about the name of the docker image used for development on the README.md file. Later, once we have a developer section in the docs, this could be added there too, but at this time the docs only contain release information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 Aqueduct is the open platform that simplifies quantum experiment management.
 
-Intuitive automation and administration features enable you to 
+Intuitive automation and administration features enable you to
 focus on consistently running and scaling up your experiments.
-Powerful APIs and a flexible plugin SDK allow you to easily 
+Powerful APIs and a flexible plugin SDK allow you to easily
 integrate with Aqueduct and streamline your workflow.
 
 This project uses the following main software stack and technologies:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,37 +76,40 @@ services:
...
+   start the server.
 
 3. Start your browser and point it to `https://localhost`.
 
 For more information please check the [documentation](https://aqueducthub.github.io/aqueductcore/).
 
+#### Development Docker Images
+
+You can find development docker images from the head of the main branch on DockerHub, `aqueducthub/aqueductcore-dev:latest`. The images are built after each commit or pull request merge to the `main` branch.
+
```
str-anger, line 90: From my experience it is always useful to have the TL;DR version of documentation, which summarises all actions into a single script. Probably something like this:

Default way to install acqueductcore on your machine is:
```
docker pull aqueducthub/aqueductcore-dev:latest
docker run aqueducthub/aqueductcore-dev:latest
```

str-anger (approved):
I have a non-blocking suggestion, but it has to validated, mu docker knowledge is an a baby level =)

README.md
samiralavi: The `docker pull` command is easy but the `docker run` command can be messy. Environments should be passed as arguments along with port configuration. That is the reason we prefer docker-compose which defines the whole config in a `yaml` file.

str-anger: But still, is there a "one, two, three" receipt, which used may follow, to start an app? Currently it is scattered through the text.

safeamiiir (approved):
Happy to have this information added into the README!

README.md
samiralavi: I see your point. This looks to me you are proposing a revision of our README and docs file. I suggest you create a PR with your requested changes then.

PR #89, opened 2024-04-30
feat: keep page info in url

safeamiiir:
Pagination and Filters are stored in the URL to be persistent during the navigations.


str-anger (commented):
I have no idea what happens here, but one change looks suspicious,
`title: ""` -> `title: null`

jatinriverlane (approved):
Good job @safeamiiir

safeamiiir:
> I have no idea what happens here, but one change looks suspicious, `title: ""` -> `title: null`

This change is because after using `searchParams` which relies on page URL that has filter/pagination info. when the field is not available, that'll return `null`.
But I'll give a try to use only one string type in order to keep it cleaner.

Issue #86, opened 2024-04-30
Backend support for Markdown preview


safeamiiir:
To provide new `text/x-markdown` as `content-type` in response of .md, .mdown, ... files.
Ref: https://mimetype.io/text/x-markdown

safeamiiir:
Closed by this #92 PR.

PR #92, opened 2024-05-02
feat(content-type): Content-Type in headers in /api/files

jatinriverlane:
# Changes
* Added util method to generate content type for files
* Implemented utility method to send content type in response headers.

### DIFF
--- a/aqueductcore/backend/routers/files.py
+++ b/aqueductcore/backend/routers/files.py
@@ -21,7 +21,6 @@ from aqueductcore.backend.services.experiment import (
     build_experiment_dir_absolute_path,
     get_experiment_by_uuid,
 )
-from aqueductcore.backend.services.utils import generate_content_type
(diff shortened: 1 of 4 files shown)


samiralavi (changes requested):
Thanks @jatinriverlane, your solution works but is very intrusive in my opinion. You are rewriting the `Content-Type` header for every file, which is done in the `FileResponse` class for you. A smarter way is to use the `media_type` argument for of the `FileResponse` for files with `.md` like this:

https://www.starlette.io/responses/#fileresponse

```python
file_extention = file_name.split(".")[-1]
media_type = "text/x-markdown" if file_extention == "md" else None
response = FileResponse(file_path, stat_result=os.stat(file_path),media_type=media_type )
```

safeamiiir (approved):

safeamiiir:
This PR closed #86 issue

PR #97, opened 2024-05-07
feat(ci): License report Github action and scripts

samiralavi:
This PR adds the required workflow and scripts to generate license reports for only the production Python and Node.js packages as markdown reports. The generated file is uploaded as an artifact to the action and it is triggered manually.


jatinriverlane (approved):
Thanks @samiralavi

safeamiiir (approved):
That looks good, easier to read.
Thanks!

PR #98, opened 2024-05-08
feat(tags): New mutation to add multiple tags to an experiment.

samiralavi:
This PR adds a new mutation to add a list of tags to an experiment. The service method is changed to support that and if there are duplicate tags in the request, an error will be returned following the exception. If the tags were already assigned to an experiment, the service safely continues and adds the new tags to the database and experiment. Unit tests and integration tests are added.

### DIFF
--- a/.github/workflows/license_report.yml
+++ b/.github/workflows/license_report.yml
@@ -14,9 +14,6 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install packages
-        run: poetry install
-
       - name: Create a markdown file with contents
         id: sets-licenses
         run: |
--- a/aqueductcore/backend/routers/graphql/inputs.py
+++ b/aqueductcore/backend/routers/graphql/inputs.py
@@ -40,6 +40,14 @@ class ExperimentTagInput:
 
 
 @strawberry.input
+class ExperimentTagsInput:
+    """Input type to add or remove tags from experiment"""
+
+    experiment_id: UUID
+    tags: List[str]
+
+
+@strawberry.input
 class ExperimentCreateInput:
     """Input type to create experiemnt"""
 
--- a/aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
+++ b/aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
@@ -7,6 +7,7 @@ from aqueductcore.backend.routers.graphql.inputs import (
     ExperimentCreateInput,
     ExperimentRemoveInput,
     ExperimentTagInput,
+    ExperimentTagsInput,
     ExperimentUpdateInput,
 )
 from aqueductcore.backend.routers.graphql.types import ExperimentData
@@ -49,11 +50,25 @@ async def add_tag_to_experiment(
 ) -> ExperimentData:
     """Add tag to experiment mutation"""
 
-    experiment = await experiment_service.add_tag_to_experiment(
+    experiment = await experiment_service.add_tags_to_experiment(
         user_info=context.user_info,
         db_session=context.db_session,
         experiment_id=experiment_tag_input.experiment_id,
-        tag=experiment_tag_input.tag,
+        tags=[experiment_tag_input.tag],
+    )
+    return experiment_model_to_node(experiment)
+
+
+async def add_tags_to_experiment(
+    context: ServerContext, experiment_tag_input: ExperimentTagsInput
+) -> ExperimentData:
+    """Add tag to experiment mutation"""
+
+    experiment = await experiment_service.add_tags_to_experiment(
+        user_info=context.user_info,
+        db_session=context.db_session,
+        experiment_id=experiment_tag_input.experiment_id,
+        tags=experiment_tag_input.tags,
     )
     return experiment_model_to_node(experiment)
 
--- a/aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
+++ b/aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
@@ -60,15 +60,15 @@ async def add_tag_to_experiment(
 
 
 async def add_tags_to_experiment(
-    context: ServerContext, experiment_tag_input: ExperimentTagsInput
+    context: ServerContext, experiment_tags_input: ExperimentTagsInput
 ) -> ExperimentData:
     """Add tag to experiment mutation"""
 
     experiment = await experiment_service.add_tags_to_experiment(
         user_info=context.user_info,
         db_session=context.db_session,
-        experiment_id=experiment_tag_input.experiment_id,
-        tags=experiment_tag_input.tags,
+        experiment_id=experiment_tags_input.experiment_id,
+        tags=experiment_tags_input.tags,
     )
     return experiment_model_to_node(experiment)
 
--- a/aqueductcore/backend/services/experiment.py
+++ b/aqueductcore/backend/services/experiment.py
@@ -390,12 +390,12 @@ async def add_tags_to_experiment(
             "Non-existing experiment with the specified ID for the user."
         )
 
-    unique_tags = set(tags)
-    if len(unique_tags) != len(tags):
+    new_tags = {tag.lower(): tag for tag in tags}
+
+    if len(set(new_tags.keys())) != len(tags):
         raise AQDValidationError("Duplicate tags are not allowed in the request.")
 
-    new_tags = {tag.lower(): tag for tag in unique_tags}
-    tag_statement = select(orm.Tag).filter(orm.Tag.key.in_(new_tags.keys())).filter()
+    tag_statement = select(orm.Tag).filter(orm.Tag.key.in_(new_tags.keys()))
     result = await db_session.execute(tag_statement)
 
     cur_db_tags = result.unique().scalars().all()
--- a/aqueductcore/backend/services/experiment.py
+++ b/aqueductcore/backend/services/experiment.py
@@ -320,8 +320,7 @@ async def create_experiment(
     )
     db_session.add(db_experiment)
 
-    for db_tag in tags_in_db:
-        db_experiment.tags.append(db_tag)
+    db_experiment.tags.extend(tags_in_db)
 
     await db_session.commit()
 
@@ -369,8 +368,8 @@ async def update_experiment(
 
 
 @validate_call(config=ConfigDict(arbitrary_types_allowed=True))
-async def add_tag_to_experiment(
-    user_info: UserInfo, db_session: AsyncSession, experiment_id: UUID, tag: ExperimentTag
+async def add_tags_to_experiment(
+    user_info: UserInfo, db_session: AsyncSession, experiment_id: UUID, tags: List[ExperimentTag]
 ) -> ExperimentRead:
     """Add tag to experiment"""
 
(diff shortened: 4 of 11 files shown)


aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
```diff
@@ -49,11 +50,25 @@ async def add_tag_to_experiment(
...
         user_info=context.user_info,
         db_session=context.db_session,
         experiment_id=experiment_tag_input.experiment_id,
-        tag=experiment_tag_input.tag,
+        tags=[experiment_tag_input.tag],
+    )
+    return experiment_model_to_node(experiment)
+
+
+async def add_tags_to_experiment(
```
str-anger, line 62: Having both is, as I understand, for compatibility?

aqueductcore/backend/routers/graphql/inputs.py
```diff
@@ -39,6 +39,14 @@ class ExperimentTagInput:
     tag: str
 
 
+@strawberry.input
+class ExperimentTagsInput:
+    """Input type to add or remove tags from experiment"""
+
+    experiment_id: UUID
```
str-anger, line 46: Is this an experiment ID? I might be mistaken, but I think we agreed on EID for "user-facing" version, and unique id (or something) for UUID.

aqueductcore/backend/services/experiment.py
```diff
@@ -369,8 +368,8 @@ async def update_experiment(
 
 
 @validate_call(config=ConfigDict(arbitrary_types_allowed=True))
-async def add_tag_to_experiment(
-    user_info: UserInfo, db_session: AsyncSession, experiment_id: UUID, tag: ExperimentTag
+async def add_tags_to_experiment(
+    user_info: UserInfo, db_session: AsyncSession, experiment_id: UUID, tags: List[ExperimentTag]
```
str-anger, line 372: May I suggest to loosen constraint to `Collection` or `Sequence`?

aqueductcore/backend/services/experiment.py
```diff
@@ -391,21 +390,25 @@ async def add_tag_to_experiment(
...
 
-    tag_key = tag.lower()
-    experiment_tags = [tag.key for tag in db_experiment.tags]
+    unique_tags = set(tags)
+    if len(unique_tags) != len(tags):
+        raise AQDValidationError("Duplicate tags are not allowed in the request.")
 
-    if tag_key in experiment_tags:
-        raise AQDDBError("DB query failed due to pre-existing tag with the Experiment.")
+    new_tags = {tag.lower(): tag for tag in unique_tags}
```
str-anger, line 397: `lower()`? Is this important? I think users may be serious about this. Also, you check uniqueness of non-lowercased tags, so you may have a key conflict, one tag will be swallowed in this generator.

tests/unittests/test_experiment_services.py
```diff
@@ -384,20 +384,109 @@ async def test_add_db_tag_to_experiment(
...
+            user_id=uuid4(), username=settings.default_username, scopes=set(UserScope)
+        ),
+        db_session=db_session,
+        experiment_id=experiments_data[0].id,
+        tags=expected_tags,
+    )
+
+    in_db_experiment_tags = [tag.name for tag in in_db_experiment.tags]
+    for item in expected_tags:
+        assert item in in_db_experiment_tags
```
str-anger, line 459:
```suggestion
    in_db_experiment_tags = {tag.name for tag in in_db_experiment.tags}
    assert expected_tags.issubset(in_db_experiment_tags)
```

tests/unittests/test_experiment_services.py
```diff
@@ -384,20 +384,109 @@ async def test_add_db_tag_to_experiment(
...
+    db_session.add(db_user)
+
+    for experiment in experiments_data:
+        db_experiment = experiment_model_to_orm(experiment)
+        db_experiment.created_by_user = db_user
+        db_session.add(db_experiment)
+
+    await db_session.commit()
+
+    expected_tags = ["test1", "test2", "test3", experiments_data[0].tags[0].name]
```
str-anger, line 447:
```suggestion
    expected_tags = {"test1", "test2", "test3", experiments_data[0].tags[0].name}
```

tests/unittests/test_experiment_services.py
```diff
@@ -384,20 +384,109 @@ async def test_add_db_tag_to_experiment(
...
+    db_session.add(db_user)
+
+    for experiment in experiments_data:
+        db_experiment = experiment_model_to_orm(experiment)
+        db_experiment.created_by_user = db_user
+        db_session.add(db_experiment)
+
+    await db_session.commit()
+
+    expected_tags = ["test1", "test1"]
```
str-anger, line 478: May I suggest to validate behaviour with `expected_tags = ["teSt1", "Test1"]`

tests/unittests/test_graphql_mutations.py
```diff
@@ -465,6 +485,71 @@ async def test_add_tag_to_experiment(
...
+                "experimentId": str(experiments_data[0].id),
+                "tags": expected_tags,
+            }
+        },
+    )
+
+    assert resp.errors is not None
+    assert resp.data is None
+
+    assert resp.errors[0].message == f"Duplicate tags are not allowed in the request."
```
str-anger, line 550:
```suggestion
    assert resp.errors[0].message == "Duplicate tags are not allowed in the request."
```

str-anger (changes requested):
Some minor suggestions and questions. In general looks good.

aqueductcore/backend/routers/graphql/inputs.py
jatinriverlane: @str-anger Yes, this is Experiment ID. We are using Alias (EID) for `get experiment` query. For all mutations, we are taking Experiment ID as experiment identifier.

str-anger: I'm then still confused with naming. Is we move away from "alias", what are the names now =)

aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
```diff
@@ -49,11 +50,25 @@ async def add_tag_to_experiment(
...
+    context: ServerContext, experiment_tag_input: ExperimentTagsInput
```
jatinriverlane, line 63: Parameter `experiment_tag_input` can be renamed to `experiment_tags_input` for uniformity.

aqueductcore/backend/services/experiment.py
```diff
@@ -391,21 +390,25 @@ async def add_tag_to_experiment(
...
+    tag_statement = select(orm.Tag).filter(orm.Tag.key.in_(new_tags.keys())).filter()
```
jatinriverlane, line 398: Having second `filter` call at the end would clear filter of `orm.Tag.key.in_(new_tags.keys())`.

jatinriverlane (approved):
Thanks @samiralavi , have added a few comments.

aqueductcore/backend/routers/graphql/mutations/experiment_mutations.py
samiralavi: Yes, it is temporary until frontend and PyAqueduct switch to the new mutation and then we will remove the old one.

aqueductcore/backend/routers/graphql/inputs.py
samiralavi: @str-anger we have not yet updated the names to the new scheme. This is happening in one go in different PRs. For now, we stick to the old naming.

aqueductcore/backend/services/experiment.py
samiralavi: I understand your point, but Pydantic when validating the argument internally does some conversion which adds overhead. The recommended approach in Pyndatic for performance is that if we know the type, it is better to `list` or `tuple`:
https://docs.pydantic.dev/latest/concepts/performance/#sequence-vs-list-or-tuple-mapping-vs-dict

aqueductcore/backend/services/experiment.py
samiralavi: We have already thought about this. The unique key of the tag is always the lower case. We also have another field, `name` that holds the value of the tag with the casing provided by the user for display purposes. Database operations happen based on the key. If you see the previous line we check if the user-provided tags are unique or not by `unique_tags = set(tags)`.

However, looking at it again, if the user provides the same tags with different casing, it will be problematic. I am going to fix it.

aqueductcore/backend/services/experiment.py
samiralavi: You are right I have missed that. The logic here has become complicated, I am trying to simplify it further.

samiralavi:
@str-anger I have pushed changes to address the bug you found along with other suggestions. Please review it and resolve the ones you believe are fine.

.github/workflows/license_report.yml
```diff
@@ -14,9 +14,6 @@ jobs:
         with:
           python-version: "3.10"
 
-      - name: Install packages
-        run: poetry install
-
```
str-anger, line 19: Anything related? Or just cleanup?

aqueductcore/backend/services/experiment.py
```diff
@@ -391,21 +390,25 @@ async def add_tag_to_experiment(
...
-    db_tag = result.scalars().first()
-    if db_tag is None:
-        db_tag = orm.Tag(name=tag, key=tag_key)
-        db_session.add(db_tag)
 
-    db_experiment.tags.append(db_tag)
+    cur_db_tags = result.unique().scalars().all()
+    for db_tag in cur_db_tags:
+        if db_tag.key in new_tags:
+            del new_tags[db_tag.key]
```
str-anger, line 404: Just wondering, I think this is not tooo serious, but if we allow to "add" an existing tag, should we keep the old one, or should we update the value?

aqueductcore/backend/services/experiment.py
```diff
@@ -391,21 +390,25 @@ async def add_tag_to_experiment(
...
+            if db_tag not in db_experiment.tags:
+                db_experiment.tags.append(db_tag)
+
+    for key, value in new_tags.items():
+        db_tag = orm.Tag(name=value, key=key)
+        db_session.add(db_tag)
+        db_experiment.tags.append(db_tag)
```
str-anger, line 411: I think these two loops (401 and 408) are basically performing 2 upsert (insert + update) calls: add non-existing tags and add tags to experiment. Again, I don't have a strong opinion, but to ensure atomic action you might probably benefit from sqlalchemy features, which has both native for postgres and universal support of upserts:
1. https://stackoverflow.com/questions/7165998/how-to-do-an-upsert-with-sqlalchemy
2. https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#insert-on-conflict-upsert

ci/unit_tests.sh
```diff
@@ -23,4 +23,3 @@ poetry install
 
 echo "Build coverage report"
 poetry run $PROJECT_ROOT/scripts/run_unit_tests.sh
-
```
str-anger, line 26: I think convention suggest to have a newline in the end?

str-anger (approved):
There are some suggestions, optional to implement

.github/workflows/license_report.yml
samiralavi: It was clean up to fix a failing workflow from prev PRs.

aqueductcore/backend/services/experiment.py
samiralavi: I know it is a bit ugly, but I couldn't think of a better way. An existing tag has two fields, the key and its display name. The key doesn't change as it is the lowercase of the tag name, but the decision to update the display name is something I intentionally didn't want to cover here, as there is a discussion about whether the user can change tags or not. I leave it for implemented later.

aqueductcore/backend/services/experiment.py
samiralavi: If I understand correctly, if something fails in the session, the whole transaction including the inserts and updates is rolled back automatically by PostgreSQL. In the second response to the first link, it is mentioned that the newer version of SQLAlchemy manages that by the session. If you think this is a real problem, please let me know.

ci/unit_tests.sh
samiralavi: It is funny the line exists locally. I have just added the VSCode config to enforce it on file save.

aqueductcore/backend/services/experiment.py
str-anger: It's not an issue to me, this is just relying more on database features, than self-made logic.

PR #99, opened 2024-05-09
feat(plugins): #91 natural support of venv's and requirements.

str-anger:
This code adds a few actions, which happen by default for each plugin at first start:
1. venv in created, if doesn't exist
2. if `requirements.txt` is present, `$venv-python -m pip install -r requirements.txt`
3. If `pyproject.toml` is present, `$venv-python -m pip install .`

$python variable is used in scripts to point to a venv python.

### DIFF
--- a/aqueductcore/backend/models/plugin.py
+++ b/aqueductcore/backend/models/plugin.py
@@ -255,6 +255,7 @@ class Plugin(BaseModel):
 
     @property
     def folder(self):
+        """ Folder with plugin specification. Raises if not initialised. """
         if self._folder is None:
             raise AQDFilesPathError(f"Plugin {self.name} folder is not known.")
         return self._folder
--- a/aqueductcore/backend/models/plugin.py
+++ b/aqueductcore/backend/models/plugin.py
@@ -120,6 +120,7 @@ class PluginFunction(BaseModel):
             self,
             plugin: Plugin,
             params: dict,
+            python: str | Path | None = None,
             timeout: int=60) -> PluginExecutionResult:
         """Passes parameters to the function code and awaits
         execution results
@@ -142,8 +143,12 @@ class PluginFunction(BaseModel):
         cwd = Path.home()
         if plugin.manifest_file:
             cwd = Path(plugin.manifest_file).parent
+
+        rich_script = self.script
+        if python:
+            rich_script = rich_script.replace("$python ", f"{python} ")
         with subprocess.Popen(
-            self.script,
+            rich_script,
             shell=True,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
@@ -203,6 +208,7 @@ class Plugin(BaseModel):
     manifest_file: Optional[str] = None
     aqueduct_key: Optional[str] = None
     params: Optional[Dict[str, Any]] = None
+    folder: Path = Path.home()
 
     @classmethod
     def from_folder(cls, path: Path) -> Plugin:
@@ -223,6 +229,7 @@ class Plugin(BaseModel):
             # TODO: somehow generate and pass it here
             plugin.aqueduct_key = ""
             plugin.validate_object()
+            plugin.folder = path
             return plugin
 
     def get_function(self, name: str) -> PluginFunction:
--- a/aqueductcore/backend/models/plugin.py
+++ b/aqueductcore/backend/models/plugin.py
@@ -208,7 +208,7 @@ class Plugin(BaseModel):
     manifest_file: Optional[str] = None
     aqueduct_key: Optional[str] = None
     params: Optional[Dict[str, Any]] = None
-    folder: Path = Path.home()
+    _folder: Optional[Path] = None
 
     @classmethod
     def from_folder(cls, path: Path) -> Plugin:
@@ -229,7 +229,7 @@ class Plugin(BaseModel):
             # TODO: somehow generate and pass it here
             plugin.aqueduct_key = ""
             plugin.validate_object()
-            plugin.folder = path
+            plugin._folder = path
             return plugin
 
     def get_function(self, name: str) -> PluginFunction:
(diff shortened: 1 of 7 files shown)


aqueductcore/backend/models/plugin.py
```diff
@@ -203,6 +208,7 @@ class Plugin(BaseModel):
     manifest_file: Optional[str] = None
     aqueduct_key: Optional[str] = None
     params: Optional[Dict[str, Any]] = None
+    folder: Path = Path.home()
```
samiralavi, line 211: Why is it initialised to the home directory of the user?

aqueductcore/backend/services/plugin_executor.py
```diff
@@ -55,22 +63,116 @@ def get_plugin(cls, plugin: str) -> Plugin:
...
+        where = cls.get_plugin(plugin).folder
+        if not where.exists() or where.is_file():
+            raise AQDValidationError(f"Plugin folder `{where}` does not exist")
+        venv_folder = where / VENV_FOLDER
+        if venv_folder.exists() and venv_folder.is_file():
+            raise AQDValidationError(f"Venv `{venv_folder}` is not a folder")
+        return venv_folder.exists()
+
+    @classmethod
+    def ensure_venv_python(cls, plugin: str) -> Path:
```
samiralavi, line 85: why ensure instead of create verb?

aqueductcore/backend/services/plugin_executor.py
```diff
@@ -55,22 +63,116 @@ def get_plugin(cls, plugin: str) -> Plugin:
...
+            Path to a python executable inside a virtual environment.
+            This path will be substituted in the `script` section
+            of manifest file, if `$python` variable is used.
+        """
+        plugin_dir = cls.get_plugin(plugin).folder
+        venv_dir = plugin_dir / VENV_FOLDER
+        python_bin = (venv_dir / PYTHON_BINARY).absolute()
+        if cls.is_venv_present(plugin=plugin):
+            return python_bin
+        venv.create(venv_dir, system_site_packages=True, with_pip=True)
```
samiralavi, line 102: I suggest disabling `system_site_packages` for better isolation. It is not the default behaviour of pip to do that unless you need system packages for a reason!

samiralavi (approved):
Thanks @str-anger, the changes look good to me. Great work.

aqueductcore/backend/models/plugin.py
str-anger: I need your advice. I want this to be mandatory, but I want to populate it later than constructor is called. I can make in optional, but then I will have to add a not None check. So, probably I just don't see an obvious solution.

aqueductcore/backend/services/plugin_executor.py
str-anger: Here the method create and populated venv if and only if it is not present. I met this before, but I agree this is not too common. You propose `create_venv_if_not_present` or similar?

aqueductcore/backend/services/plugin_executor.py
str-anger: deal

aqueductcore/backend/models/plugin.py
samiralavi: I guess what you described means that it is not mandatory at the time the objects get created. The definition of mandatory and optional are relevant to the time the objects get created via the constructor. My advice is to make it mandatory and at the place the object is getting created, the value should be decided.

aqueductcore/backend/services/plugin_executor.py
```diff
@@ -21,6 +23,12 @@
     get_experiment_by_alias,
 )
 
+
+VENV_FOLDER = ".aqueduct-plugin-venv"
+PYTHON_BINARY = "bin/python"
+EXEC_TIMEOUT = 600
```
safeamiiir, line 29: This bit I wasn't sure about, wanted to ask more.
So basically if there's not venv created by the user we'll craete one for them at this address?
is `PYTHON_BINARY = "bin/python"` same in all venvs?

safeamiiir (approved):
Thanks Stas! As far as I'm familiar with python vens nothing specific to add

aqueductcore/backend/services/plugin_executor.py
str-anger: >  if there's not venv created by the user we'll craete one for them at this address?

If there is no venv present (users do not create them themselves), we will initialise one at the first execution. The place for a venv - inside a plugin folder

> is PYTHON_BINARY = "bin/python" same in all venvs?

venv is a convention. This is a folder with some specific files inside. venv is a technology for python virtual environments, so the only thing I can guarantee is that there is a python binary in `bin` folder, and activation script next to it =)

safeamiiir: Yeah I see.
Fair enough.
Thanks!

aqueductcore/backend/services/plugin_executor.py
```diff
@@ -55,22 +63,117 @@ def get_plugin(cls, plugin: str) -> Plugin:
...
+            of manifest file, if `$python` variable is used.
+        """
+        plugin_dir = cls.get_plugin(plugin).folder
+        venv_dir = plugin_dir / VENV_FOLDER
+        python_bin = (venv_dir / PYTHON_BINARY).absolute()
+        if cls.is_venv_present(plugin=plugin):
+            return python_bin
+        venv.create(venv_dir, system_site_packages=False, with_pip=True)
+        cls.try_install_requirements_txt(plugin=plugin, python=python_bin)
+        cls.try_install_pyproject_toml(plugin=plugin, python=python_bin)
```
gabrielg42, line 105: What happens if both are provided?

str-anger: Both are installed. requirements first, and then package with its dependencies

PR #113, opened 2024-05-14
feat(plugins): merging basic backend plugin functionality into main

str-anger:
Plugin support from at the backend, includes core plugin support, does not include distribution changes and documentation.


aqueductcore/backend/settings.py
```diff
@@ -50,4 +50,8 @@ class Settings(BaseSettings):
     keycloak_client_id: Optional[str] = None
     """Keycloak client ID (used in Aqueduct Pro only)"""
 
+    plugins_dir_path: str = "/workspace/plugins"
```
samiralavi, line 53: The default value smells.

samiralavi: It should be optional for the case when the user doesn't have any plugins or before releasing the plugins feature.

samiralavi (approved):
Thanks @str-anger, all is fine. Only care should be taken regarding the environment variable as it affects the release before the feature gets completed. If you think before the next release we can finish the feature, then I say leave it.

aqueductcore/backend/models/plugin.py
```diff
@@ -0,0 +1,261 @@
...
+                return str_value
+            except Exception as exc:
+                raise AQDValidationError(f"{str_value} is not float") from exc
+
+        if self.data_type == SupportedTypes.BOOL:
+            if str_value.lower() in ("true", "1"):
+                return "1"
+            if str_value.lower() in ("false", "0"):
+                return "0"
+            raise AQDValidationError(f"{value} is not a valid boo.")
```
safeamiiir, line 88: the letter L is missed in the end.

safeamiiir (approved):
Thanks Stat!
Let's have it in main

PR #114, opened 2024-05-14
feat(drawer): make more exportable functions

safeamiiir:
Make code more reusable as a dependency

### DIFF
--- a/aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
+++ b/aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
@@ -1,29 +1,29 @@
 import { Link, useLocation, useSearchParams } from "react-router-dom";
+import { PropsWithChildren, ReactNode, useState } from "react";
 import DescriptionIcon from '@mui/icons-material/Description';
 import { ExpandLess, ExpandMore } from "@mui/icons-material";
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import SettingsIcon from "@mui/icons-material/Settings";
 import ScienceIcon from "@mui/icons-material/Science";
-import { PropsWithChildren, useState } from "react";
 import MenuIcon from "@mui/icons-material/Menu";
 import { Box } from "@mui/system";
 import {
-  AppBar,
-  Badge,
-  BadgeProps,
-  Collapse,
-  Drawer,
-  Grid,
-  IconButton,
-  List,
-  ListItem,
   ListItemButton,
-  ListItemIcon,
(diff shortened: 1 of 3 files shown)


aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
```diff
@@ -1,29 +1,29 @@
 import { Link, useLocation, useSearchParams } from "react-router-dom";
+import { PropsWithChildren, ReactNode, useState } from "react";
```
str-anger, line 2: What's the reason to move a string if you add an item to import? I see they are ordered by length, which looks like the weirdest criterion you might imagine.

aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
```diff
@@ -1,29 +1,29 @@
...
 import DescriptionIcon from '@mui/icons-material/Description';
 import { ExpandLess, ExpandMore } from "@mui/icons-material";
 import OpenInNewIcon from '@mui/icons-material/OpenInNew';
 import SettingsIcon from "@mui/icons-material/Settings";
 import ScienceIcon from "@mui/icons-material/Science";
-import { PropsWithChildren, useState } from "react";
 import MenuIcon from "@mui/icons-material/Menu";
 import { Box } from "@mui/system";
 import {
```
str-anger, line 10: Some unexplainable reordering =)

str-anger (commented):
Each next PR looks clearer for me, but I still don't see any power to approve frontend PRs =)

samiralavi (approved):

aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
safeamiiir: Well it's not just the length and I don't believe it's weirdest as it's just to make it more readable.
First section includes third party libraries and then internal ones. But within each section order is by length and this needs to be done through a linter (even with a different criteria) which I'm putting off for a while 😁 But will do in a separated PR because it's really annoying.

aqueductcore/frontend/src/components/templates/drawerLayout/index.tsx
safeamiiir: I was looking for options and I'll config [this one](https://github.com/azat-io/eslint-plugin-perfectionist) in the next PR.
Thanks for pushing it 😁

jatinriverlane (approved):
Thanks @safeamiiir

PR #116, opened 2024-05-16
feat(cli): CLI Exporter module for exporting instance data into tarfile

samiralavi:
This PR is the first one that implements the `Exporter` module as a part of the CLI application to export Aqueduct instance data (metadata and experiments files) into a tar file. The chosen compression method is gzip. Basic unit tests are added, but they will be improved in future PRs related to CLI and import functionality.

### DIFF
--- a/aqueductcore/cli/export.py
+++ b/aqueductcore/cli/export.py
@@ -1,0 +1,139 @@
+"""Aqueduct module for exporting data from instance."""
+
+from __future__ import annotations
+
+import errno
+import os.path
+import tarfile
+from io import BytesIO
+from typing import Any, Callable, Optional
+
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from aqueductcore import __version__
+from aqueductcore.backend.errors import AQDFilesPathError
+from aqueductcore.backend.models import orm
+from aqueductcore.cli.models import AqueductData, AqueductVariant, Experiment, Tag, User
+from aqueductcore.cli.session import Session
+
+
+class Exporter:
+    """Aqueduct exporter class."""
+
+    EXPERIMENTS_BASE_DIR_NAME = "experiments_files"
+
+    @classmethod
+    def export_experiments_metadata(
+        cls,
+        db_session: Session,
+    ) -> AqueductData:
+        """Export instance data as an object.
+
+        Args:
+            db_session: Database session to load metadata from.
+
+        Returns:
+            Instance metadata as an AqueductData object.
+
+        """
+        statement = select(orm.User)
+
+        result = db_session.execute(statement)
+
+        data = AqueductData(version=__version__, variant=AqueductVariant.CORE, users=[])
+        for user in result.unique().scalars().all():
+            user_data = User(uid=user.id, username=user.username, experiments=[])
+            for experiment in user.experiments:
+                user_data.experiments.append(
+                    Experiment(
+                        eid=experiment.alias,
+                        title=experiment.title,
+                        description=experiment.description,
+                        created_at=experiment.created_at,
+                        updated_at=experiment.updated_at,
+                        tags=[Tag(key=item.key, name=item.name) for item in experiment.tags],
+                    )
+                )
+            data.users.append(user_data)
+
+        return data
+
+    @classmethod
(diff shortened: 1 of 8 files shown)


aqueductcore/cli/export.py
```diff
@@ -0,0 +1,138 @@
...
+from aqueductcore import __version__
+from aqueductcore.backend.errors import AQDFilesPathError
+from aqueductcore.backend.models import orm
+from aqueductcore.cli.models import AqueductData, AqueductVariant, Experiment, Tag, User
+
+
+class Exporter:
+    """Aqueduct exporter class."""
+
+    EXPERIMENTS_BASE_DIR_NAME = "experiments_files"
```
str-anger, line 23: Wow-wow, constant dir name detected! ❗

aqueductcore/cli/models.py
```diff
@@ -0,0 +1,51 @@
...
+
+from datetime import datetime
+from enum import Enum
+from typing import List, Optional
+from uuid import UUID
+
+from pydantic import BaseModel
+
+
+class Tag(BaseModel):
```
str-anger, line 13: Is this similar to ORM class? May you reuse that?

aqueductcore/cli/export.py
```diff
@@ -0,0 +1,138 @@
...
+            Size of the directory including its files recursively in bytes.
+
+        """
+        total = 0
+        with os.scandir(experiments_dir) as iterator:
+            for entry in iterator:
+                if entry.is_file():
+                    total += entry.stat().st_size
+                elif entry.is_dir():
+                    total += Exporter._get_dir_size(entry.path)
```
str-anger, line 78:
```suggestion
                    total += cls._get_dir_size(entry.path)
```

aqueductcore/cli/export.py
```diff
@@ -0,0 +1,138 @@
...
+        return total
+
+    @classmethod
+    def export_artifact(
+        cls,
+        metadata: bytes,
+        output_fileobj: BytesIO,
+        metadata_filename="metadata.json",
+        experiments_root: Optional[str] = None,
+        progress: Optional[Callable[[int], Any]] = None,
```
str-anger, line 88: This is a CLI. Maybe you can just use `tqdm` inside?

aqueductcore/cli/session.py
```diff
@@ -0,0 +1,29 @@
...
+from contextlib import contextmanager
+from typing import Generator
+
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session, sessionmaker
+
+from aqueductcore.backend.settings import settings
+
+sync_engine = create_engine(
+    f"postgresql+psycopg2://{settings.postgres_username}:"
```
str-anger, line 14: Whatever you claim about "use any database", your code is written with respect to postgres =)

str-anger (changes requested):
Intrigued to see new PRs. Some minor requests here.

aqueductcore/cli/export.py
samiralavi: In fact, we need a constant directory name in the tar file to copy all of the experiment files into. The location of this constant could be seen outside this class, however, as this data is related to the exporter, it is better to be stored here, unless you have a better suggestion.

aqueductcore/cli/models.py
samiralavi: I intentionally wanted to replicate the type to completely isolate the modules from each other as each datatype is used for a different purpose. The cost is a bit of duplication but with a better object structure.

aqueductcore/cli/session.py
samiralavi: Ah this is just the connection string, you can just change the driver, but everything else stays the same. The main thing stopping me from adding support for multiple to Aqueduct is that I don't know a good way to expose this to the users.

aqueductcore/cli/export.py
str-anger: This one is obeying access modifier, which is better than not obeying.

str-anger (approved):

samiralavi:
I am closing this PR so you can review both import and export together in one PR: https://github.com/AqueductHub/aqueductcore/pull/119

PR #118, opened 2024-05-20
feat(automatic-release): Automated release for aqueduct core

jatinriverlane:
# Changes
* update trigger from workflow dispatch to cron
* add condition of deployment to confirm with the environment variable of ALLOW_AUTOMATED_RELEASE
* add generateReleaseNotes as True to automatically add release notes


.github/workflows/release.yaml
```diff
@@ -1,55 +1,76 @@
 name: Release Aqueductcore
 
 on:
-  workflow_dispatch:
```
samiralavi, line 4: Please make sure the `workflow_dispatch` is there. We might want to still release manually instead of waiting for two weeks.

.github/workflows/release.yaml
```diff
@@ -1,55 +1,76 @@
...
-        description: "Aqueduct Version"
-        required: true
-        type: string
-        default: "MAJOR.MINOR.PATCH"
+  schedule:
+    - cron: "30 11 * * 1"
 jobs:
   build:
     runs-on: ubuntu-20.04
+    if: ${{ env.ALLOW_AUTOMATED_RELEASE == 'true' }}
```
samiralavi, line 9: You need to add a condition for the manual trigger to bypass this.

.github/workflows/release.yaml
```diff
@@ -1,55 +1,76 @@
...
       contents: write
     steps:
+      - name: Confirm action execution
+        id: is_release_allowed_check
+        run: |
+          echo "is_release_allowed=$(( $(date +%V) % 2 == 0 ))" >> $GITHUB_OUTPUT
+
       - name: Checkout repository
         uses: actions/checkout@v3
+        if: ${{ steps.is_release_allowed_check.outputs.is_release_allowed }}
```
samiralavi, line 20: Instead of adding this check to every task of the `build` job, you can create a new job before `build` and make sure the `build` depends on the job. If the job gets then the dependant jobs don't get executed. Think of it like a pipeline.

.github/workflows/release.yaml
samiralavi: This also helps us to change the major and minor version numbers ourselves, but let the CI increment only the patch version.

.github/workflows/release.yaml
```diff
@@ -1,55 +1,76 @@
...
+          echo "version=$NEXT_VERSION" >> $GITHUB_OUTPUT
+
+          if $(( $(date +%U) % 2 == 0 ))  {
+            exit 0
+          }
 
       - name: Build Release Images
-        run: ci/build_docker_image.sh -i aqueductcore -p true -t ${{ inputs.version }}
+        if: ${{ steps.is_release_allowed_check.outputs.is_release_allowed }}
+        run: ci/build_docker_image.sh -i aqueductcore -p true -t ${{ steps.next_version.outputs.version }}
```
samiralavi, line 52: for testing purposes, I recommend changing this to `aqueductcore-dev` until you made sure everything is working fine.

samiralavi (changes requested):
Generally looks good but there are points to improve.

.github/workflows/release.yaml
jatinriverlane: @samiralavi Shouldn't we have different workflows for `automatic release triggered by cron having no input` and `manual release triggered manually with version input`. Although having different workflows will have some code duplication, but we won't have to apply multiple if else conditions in case of single workflow. Having them differently will be less error prone as well. 

What are your views on this ?

samiralavi (approved):
Thanks @jatinriverlane, this is what I call a clean approach.

safeamiiir (approved):
Thanks Jatin!

I checked the ncipollo/release-action and it'll fail if we don't have changes from the previous release based on `allowUpdates` input and by default it's false and it'll fail the release if it already exists.

PR #119, opened 2024-05-21
feat(cli): Modules to import/export metadata and experiment files into the Aqueduct instance

samiralavi:
This PR adds the import and export modules to the CLI application for importing aqueduct artifacts that are previously exported. Tests are added to check the functionality. The following describes the importing behaviour in certain scenarios:

- When importing, if there is an existing experiment with the same experiment ID (aka alias), the import fails, outputting the conflicting experiments.
- When importing, if there are already existing conflicting tags, the existing ones will be used to be assigned to the imported experiment. This means the original display case (name field) will be preserved.
- When importing, if the are already existing conflicting users with the same user ID, the original username will be kept and the experiments will be assigned to the existing user. Conflicting user IDs means they are the same user as the user ID is a UUID.
- Due to the way the Linux file system and archiving works, the file’s last updated time will be reset to the time the files are extracted from the archive. If the original datetime is required by the user, the volume should be moved by the system administrator instead.

About the export module, the chosen compression method is gzip. Basic unit tests are added, but they will be improved in future PRs related to CLI and import functionality. The rest of the code is self-explanatory.

### DIFF
--- a/aqueductcore/cli/exporter.py
+++ b/aqueductcore/cli/exporter.py
@@ -92,9 +92,8 @@ class Exporter:
         gzip compression.
 
         Args:
-            metadata: Aqueduct metadata as bytes.
-            output_fileobj: Output file object for the generated tar file.
-            metadata_filename: Metadata file name to get into the generated tar file.
+            metadata_bytes: Aqueduct metadata encoded as bytes.
+            tar: Tar file to generate the archive.
             experiments_root: Experiments rood directory of the Aqueduct instance.
             progress: Call back with processed data information to show progress.
 
--- a/aqueductcore/cli/exporter.py
+++ b/aqueductcore/cli/exporter.py
@@ -59,7 +59,7 @@ class Exporter:
         return data
 
     @classmethod
-    def get_dir_size(cls, experiments_dir: str) -> int:
+    def get_size(cls, path: str) -> int:
         """Get the directory size in bytes given the path.
 
         Args:
@@ -69,13 +69,15 @@ class Exporter:
             Size of the directory including its files recursively in bytes.
 
         """
+        if os.path.isfile(path):
+            return os.path.getsize(path)
         total = 0
-        with os.scandir(experiments_dir) as iterator:
+        with os.scandir(path) as iterator:
             for entry in iterator:
                 if entry.is_file():
                     total += entry.stat().st_size
                 elif entry.is_dir():
-                    total += cls.get_dir_size(entry.path)
+                    total += cls.get_size(entry.path)
         return total
 
     @classmethod
@@ -108,18 +110,10 @@ class Exporter:
         if experiments_root:
             with os.scandir(experiments_root) as dir_iterator:
                 for entry in dir_iterator:
-                    entry_size = 0
-                    if entry.is_file(follow_symlinks=False):
-                        entry_size = entry.stat().st_size
-                        tar.add(
-                            name=entry.path,
-                            arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
-                        )
-                    elif entry.is_dir(follow_symlinks=False):
-                        entry_size = cls.get_dir_size(entry.path)
-                        tar.add(
-                            name=entry.path,
-                            arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
-                        )
+                    entry_size = cls.get_size(entry.path)
+                    tar.add(
+                        name=entry.path,
+                        arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
+                    )
                     if progress:
                         progress(entry_size)
--- a/aqueductcore/cli/exporter.py
+++ b/aqueductcore/cli/exporter.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import errno
 import os.path
 from io import BytesIO
 from tarfile import TarFile, TarInfo
@@ -12,7 +11,6 @@ from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from aqueductcore import __version__
-from aqueductcore.backend.errors import AQDFilesPathError
 from aqueductcore.backend.models import orm
 from aqueductcore.cli.models import AqueductData, AqueductVariant, Experiment, Tag, User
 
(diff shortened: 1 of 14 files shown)


aqueductcore/cli/exporter.py
```diff
@@ -0,0 +1,135 @@
...
+                    for entry in dir_iterator:
+                        entry_size = 0
+                        if entry.is_file(follow_symlinks=False):
+                            entry_size = entry.stat().st_size
+                            tar.add(
+                                name=entry.path,
+                                arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
+                            )
+                        elif entry.is_dir(follow_symlinks=False):
+                            entry_size = cls._get_dir_size(entry.path)
```
str-anger, line 123: It looks like you call dir size multiple times in recursion, so probably _get_dir_size() may benefit from `@cache` decorator. Also, it you allow this method to return size for files as well, your code will become clearer, as you will call it once outside if statement. As I understand, branching here depends on size computation only.

aqueductcore/cli/importer.py
```diff
@@ -0,0 +1,149 @@
...
+
+        cur_users_statement = select(orm.User).where(
+            orm.User.id.in_([item.uuid for item in metadata.users])
+        )
+        cur_db_users = db_session.execute(cur_users_statement).scalars().all()
+        cur_db_users_dict = {item.id: item for item in cur_db_users}
+
+        metadata_experiments: List[Experiment] = []
+        for user in metadata.users:
+            metadata_experiments.extend(user.experiments)
```
str-anger, line 84: I did not find where you use `get_conflicting_experiments`, but my opinion is that import should be as atomic as possible. Which means (to me) that if there is a conflict (from the definition given in PR), the whole migration should not happen. What I see here, that if version is compatible, you will add some users, but experiments may raise and will be partially inserted. I may be wrong, if the whole method in wrapped into a DB transaction

tests/unittests/cli/test_exporter.py
```diff
@@ -0,0 +1,114 @@
...
+            )
+
+        export_tarfile.seek(0)
+        with tarfile.open(fileobj=export_tarfile, mode="r:gz") as tar:
+            for member in tar:
+                if member.isfile():
+                    assert os.path.join(tmpdirname, member.name) in list(test_files.keys())
+                    file = tar.extractfile(member)
+                    assert file is not None
+                    assert file.read() == test_files[os.path.join(tmpdirname, member.name)]
```
str-anger, line 114: It would be nice to have tests for `pytest.raises(..., match="abc")` scenarios.

aqueductcore/cli/exporter.py
samiralavi: I didn't know about `@cache`. Thanks for the suggestion. I looked at it and I see your point, however, it is only looping through the experiments directory individually and it never reads that directory again. You can see that `dir_iterator` is iterated over only once, so we won't get any performance improvement here, unless there is something else I have missed.

aqueductcore/cli/importer.py
samiralavi: That function will be called by the CLI app in my next PR. I have renamed the functions so they are only doing one thing at a time. Rest assured that this will be called before the import.

tests/unittests/cli/test_exporter.py
samiralavi: I have moved file handling and their corresponding exception handling to the CLI app in my next PR as we need to catch them there to show custom messages. Another change is that the file resource management happens in the CLI app so exceptions will be tackled there.

aqueductcore/cli/exporter.py
str-anger: I'd say, joining the method into something like `get_fs_object_size` will remove unnecessary branching here: you get the size and then you put the object into a tar.
Also, what is potential "not dir" and "not file"? you have `elif` there.

aqueductcore/cli/exporter.py
```diff
@@ -0,0 +1,125 @@
...
+                        updated_at=experiment.updated_at,
+                        tags=[Tag(key=item.key, name=item.name) for item in experiment.tags],
+                    )
+                )
+            data.users.append(user_data)
+
+        return data
+
+    @classmethod
+    def get_dir_size(cls, experiments_dir: str) -> int:
```
str-anger, line 62: I still feel like this may be reduced to:
```
def get_size(cls, obj: str) -> int:
        if os.path.isfile(obj):
            return os.path.getsize(obj)
        total = 0
        with os.scandir(obj) as iterator:
            for entry in iterator:
                if entry.is_file():
                    total += entry.stat().st_size
                elif entry.is_dir():
                    total += cls.get_dir_size(entry.path)
        return total
```

aqueductcore/cli/exporter.py
```diff
@@ -0,0 +1,125 @@
...
+                        tar.add(
+                            name=entry.path,
+                            arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
+                        )
+                    elif entry.is_dir(follow_symlinks=False):
+                        entry_size = cls.get_dir_size(entry.path)
+                        tar.add(
+                            name=entry.path,
+                            arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
+                        )
```
str-anger, line 123: And if you do a `get_size` then:

```suggestion
                    entry_size = cls.get_size(entry.path)
                    tar.add(
                        name=entry.path,
                        arcname=os.path.join(cls.EXPERIMENTS_BASE_DIR_NAME, entry.name),
                    )
```

BTW, what is the reason NOT to follow symlinks? if there is a file in a folder, visible in the interface, why would you exclude it from export?

str-anger (approved):
My comments are more into a code clarity, and one minor comment of symlinks. The rest is ok to go.

aqueductcore/cli/exporter.py
samiralavi: Symlinks don't work when mounting a directory from the host to the container. Also, we are in control of files in the experiments directory and there are no chances that symlinks get generated.

samiralavi:
Thanks @str-anger for the code snippet. It improves the size function and its code readability. I appreciate it and I have changed the code.

tests/unittests/cli/test_importer.py
```diff
@@ -0,0 +1,142 @@
+# pylint: skip-file
+
+import os
+import random
+import tarfile
+from datetime import datetime
+from io import BytesIO
+from tempfile import TemporaryDirectory
+from typing import Dict, List
```
safeamiiir, line 9: Looks like `Dict` is not used

tests/unittests/cli/test_exporter.py
```diff
@@ -0,0 +1,114 @@
...
+from aqueductcore.cli.exporter import Exporter
+from aqueductcore.cli.models import AqueductData, AqueductVariant, Experiment, Tag, User
+
+
+def test_export_experiments_metadata(db_session: Session, experiments_data: List[ExperimentCreate]):
+
+    expected_metadata = AqueductData(
+        version=aqueductcore.__version__, variant=AqueductVariant.CORE, users=[]
+    )
+    for _ in range(10):  ## TODO
```
safeamiiir, line 26: `TODO` is resolved? maybe we can remove it.

safeamiiir (approved):
Thanks Amir!
some minor suggestions only.

tests/unittests/cli/test_exporter.py
samiralavi: Yeah I missed that.

PR #124, opened 2024-05-23
feat(cli): Aqueduct CLI application with Import/Export function

samiralavi:
This PR implements the CLI application for import-export functionality. The library used to implement the CLI is Typer which is mature and fits our technology stack based on Pydantic: https://typer.tiangolo.com/. I was aware of "tqdm" too, but this seems to let me achieve more with less complexity when developing the CLI with good functionality. The CLI interface is:

image


samiralavi:
I am going to add a unit test for the import part with the new field in the `metadata.json` for the export datetime in UTC.

aqueductcore/backend/services/plugin_executor.py
```diff
@@ -49,17 +45,17 @@ def list_plugins(cls) -> List[Plugin]:
                 except AQDValidationError as err:
                     logging.error(err)
                     # TODO: raise again here?
-        return result
+        return sorted(result, key=lambda plugin: plugin.name)
```
samiralavi, line 48: The main line changed in this file is this one, due to a failing test for the plugins.

aqueductcore/cli/commands.py
```diff
@@ -0,0 +1,130 @@
...
+    if not path.name.endswith(tar_suffix):
+        err_console.print(
+            f"[bold yellow]Warning[/bold yellow]: Archive path '{path}'"
+            f" doesn't end with '{tar_suffix}'."
+        )
+
+    with path.open(mode="wb") as file:
+        console.print(f"Adding experiments' metadata to '{file.name}'...")
+        with get_session(sync_engine) as db_session:
+            orm.Base.metadata.create_all(sync_engine)  # create relations if database is empty
```
jatinriverlane, line 40: Do we need this line here? The tables should already be created during the installation steps. Although this won't be having any affect if tables are already present, we can just rethink of having this.

samiralavi: Hmm yeah it might not be needed as the FastAPI during startup (not installation time) inits the database. I see it as a safety measure in a very rare scenario that the user has not run Aqueduct but prepared the environment. It doesn't have any side effect.

aqueductcore/cli/commands.py
```diff
@@ -0,0 +1,130 @@
...
+            console.print(
+                f"Importing experiments' files from '{file.name}'"
+                f" to '{settings.experiments_dir_path}'..."
+            )
+            with typer.progressbar(length=total, label="Experiments") as progress:
+                Importer.import_experiment_files(
+                    tar,
+                    str(settings.experiments_dir_path),
+                    progress=progress.update,
+                )
```
jatinriverlane, line 128: One case here that we can cover is the cancellation of the import process while it is working.

If while importing the command gets stuck, for eg. `Disk full`, `system shutdown`, etc., the user will end up with partial data in the system with some files. We can commit the data to database after all the files were imported.

jatinriverlane (approved):
Very good job with the CLI @samiralavi , Thanks

safeamiiir (approved):
Thanks Amir!

PR #133, opened 2024-05-29
fix(chore): restructure files

safeamiiir:
- Restructure API folders
- And a small change for the .env.dev url


jatinriverlane (approved):
Thanks @safeamiiir

PR #134, opened 2024-05-29
feat(extension): Add execution functionality in the UI

safeamiiir:
Finally these are are ready!
- Renaming
-- plugin -> extension
-- function -> action
- Extension execution is handled in the frontend

*todo: refresh is still needed to see the result and it's not selecting the result file automatically


samiralavi (approved):

aqueductcore/frontend/src/constants/constants.ts
```diff
@@ -11,3 +11,15 @@ export const experimentRecordsRowsPerPageOptions = [10, 25, 100];
...
+export const ExtensionParameterDataTypes = {
+    STR: "str",
+    INT: "int",
+    FLOAT: "float",
+    EXPERIMENT: "experiment",
+    TEXTAREA: "textarea",
+    BOOL: "bool",
+    SELECT: "select",
+    FILE: "file"
+}
```
jatinriverlane, line 25: Can't we use typescript enum instead of JSON here ?

aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
```diff
@@ -108,24 +167,34 @@ function ExtensionModal({ isOpen, handleClose, selectedExtension }: ExtensionMod
...
+                        <ActionForm selectedAction={selectedAction} inputParams={inputParams} setInputParams={setInputParams} />
+                    </ModalStepGrid> : null}
+                    <ModalFooter>
+                        <Box sx={{ display: 'flex', justifyContent: 'flex-end' }}>
+                            <Button
+                                size="small"
+                                variant="contained"
+                                onClick={() => handleExecuteExtension()}
+                                title='run_extension'
+                            >{loading ? 'running' : 'Run Extention'}</Button>
```
jatinriverlane, line 191: How about instead of just the string "running", I guess we can add the <CircularProgress /> from mui.

jatinriverlane: Or, this can actually be handled in some next PR

jatinriverlane (approved):
Great work @safeamiiir , all changes are accurate.

aqueductcore/frontend/src/constants/constants.ts
safeamiiir: I was actually not sure about using which, I found this two reasons not to use `Enum` as we're also in need of these in `Aqueduct-pro` 
- https://www.typescriptlang.org/docs/handbook/enums.html#const-enum-pitfalls
- https://youtu.be/jjMbPt_H3RQ?si=kr_EPQJp9IIfuv6a

aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
safeamiiir: Sorry this part was missed.
Fixed now

PR #136, opened 2024-06-03
feat(docker): Enable SSH access to the container via Azure

samiralavi:
Following the discussions around accessing container, to enable access for Azure web app into the container, ssh configuration following Azure docs is added to the container:
https://learn.microsoft.com/en-us/azure/app-service/configure-custom-container?pivots=container-linux&tabs=debian#enable-ssh
https://learn.microsoft.com/en-us/azure/app-service/configure-linux-open-ssh-session?pivots=container-linux

I know it is not ideal to have it in the release container and probably in future we create a new docker image for Azure hosted web apps.


safeamiiir (approved):
Looks good to me
Thanks Amir!

jatinriverlane (approved):
Thanks @samiralavi

PR #143, opened 2024-06-05
feat(backend)!: Rename id to uuid and alias to eid everywhere

samiralavi:
This big PR is all about renaming id to uuid and alias to eid. I guess my eyes's number increased by 1 as I tried to find out every usage and rename them. VSCode Python auto-rename function helped a lot though.


aqueductcore/backend/models/experiment.py
```diff
@@ -29,23 +29,22 @@ class TagRead(TagBase):
...
 
-    alias: str
+    eid: str
     tags: List[TagCreate]
 
 
 class ExperimentRead(ExperimentBase):
     """Model fields to read an experiment"""
 
-    id: UUID
```
str-anger, line 47: Is this intentionally you drop UUID?

aqueductcore/backend/models/experiment.py
```diff
@@ -59,5 +58,5 @@ class ExperimentUpdate(ExperimentBase):
 class UserBase(AQDModel):
     """Base model for User"""
 
-    user_id: UUID
+    uuid: UUID
```
str-anger, line 61: in `UserInfo` you use `user_uuid` name.

aqueductcore/backend/models/plugin.py
```diff
@@ -91,10 +91,10 @@ def validate_value(self, value: str) -> str:
             try:
                 prefix, postfix = value.split("-")
                 if not prefix.isdecimal() or not postfix.isalnum():
-                    raise AQDValidationError("Experiment alias has wrong format.")
+                    raise AQDValidationError("Experiment EID has wrong format.")
```
str-anger, line 94: EID is already Experiment ID :)

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -128,7 +131,7 @@ async def execute_plugin(
         exp_parameter = function_object.get_default_experiment_parameter()
         if exp_parameter is None:
             raise AQDValidationError(f"Function {plugin}/{function} has no experiment parameters")
-        exp_id = dict_params[exp_parameter.name]
+        eid = dict_params[exp_parameter.name]
```
str-anger, line 134: Hmmmm, but okay =)

aqueductcore/backend/services/experiment.py
```diff
@@ -245,7 +245,7 @@ async def get_experiment_files(
     return file_names
 
 
-EXPERIMENT_ALIAS_PATTERN = r"^(19[0-9]{2}|2[0-9]{3})(0[1-9]|1[012])([123]0|[012][1-9]|31)-(\d+)$"
+EXPERIMENT_EID_PATTERN = r"^(19[0-9]{2}|2[0-9]{3})(0[1-9]|1[012])([123]0|[012][1-9]|31)-(\d+)$"
```
str-anger, line 248: EID is already Experiment ID

str-anger (approved):
I'm probably ok. Some minor comments of naming

aqueductcore/backend/models/experiment.py
samiralavi: Yes, it was a kind of bug as the `id` was already defined in the base class.

aqueductcore/backend/models/experiment.py
samiralavi: yeah now I look at it, it is better to change that to `uuid` instead.

aqueductcore/backend/models/plugin.py
samiralavi: I know it is repetitive but I wanted to make sure everywhere we are talking about the ID, the term EID and UUID is clearly used to denote that.

aqueductcore/backend/services/experiment.py
samiralavi: Yeah, I didn't want to use `ID` on its own without any prefix anywhere to make sure the reference is clear

jatinriverlane (approved):
Good job Amir

PR #154, opened 2024-06-13
feat(files): open log file when extension is executed

safeamiiir:
- Select the file from the file list and open it in the content viewer

### DIFF
--- a/aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
+++ b/aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
@@ -1,19 +1,21 @@
 import { Box, Button, CircularProgress, Grid, Modal, Typography, styled } from "@mui/material"
 import ChevronRightIcon from '@mui/icons-material/ChevronRight';
 import AutoAwesomeIcon from '@mui/icons-material/AutoAwesome';
+import { useContext, useEffect, useState } from "react";
 import CloseIcon from '@mui/icons-material/Close';
 import { useParams } from "react-router-dom";
-import { useEffect, useState } from "react";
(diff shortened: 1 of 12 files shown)


str-anger (approved):

aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
```diff
@@ -112,8 +115,12 @@ function ExtensionModal({ isOpen, handleClose, selectedExtension }: ExtensionMod
         )
     }, [selectedExtension, selectedAction])
 
-    function handleOnCompletedExtensionExecution(executeExtension: EXECUTE_EXTENSION_TYPE) {
+    async function handleOnCompletedExtensionExecution(executeExtension: EXECUTE_EXTENSION_TYPE) {
         handleClose()
```
jatinriverlane, line 119: Currently, on successful execution of extension, the extension window gets closed, but the data in the form stays in the modal. i.e. after execution, when we open the extension modal again, the previous data is still present. 

Logic for clearing the fields can be added here.

jatinriverlane (approved):
Thanks @safeamiiir

aqueductcore/frontend/src/components/organisms/ExtensionModal/index.tsx
safeamiiir: You're right.
That's actually part of the enhancement ticket, that we can move it to a new one to be fair! could you please do that?
I just didn't want to touch on that in this PR.

safeamiiir: will be handled here #162

PR #156, opened 2024-06-17
fix(tags): add shouldIncludeTags in the init phase

safeamiiir:
fix shouldIncludeTags in the start initialisation of the page.


samiralavi (approved):

jatinriverlane (approved):
Thanks @safeamiiir

safeamiiir:
> Is there a way to make it less single-liner?
Thanks for poking me on this, I made it in a function

> Also, brackets are not useless here:

Brackets are because it accepts an array, and it's a one item array in this case.

PR #160, opened 2024-06-18
feat(extensions): replacing extension folder with the submodule

str-anger:
This PR replaces the folder with extensions with a separate repo. It also takes care about artefacts which are in the new repo, but were not handled properly before.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "extensions"]
```
safeamiiir, line 1: I guess it might be better to call them _external_ to make it more separated from the rest.

```suggestion
[submodule "external/extensions"]
```

We might also need to modify extensions dir path:
```env
EXTENSIONS_DIR_PATH=/workspace/external/extensions
```

safeamiiir (approved):
Very nice separation!
Thanks Stas

.gitmodules
str-anger: They a replacing the folder. There will be no extension code in the the repo any more. @safeamiiir  May you please motivate this change?

safeamiiir: Yeah I get that.
It's more about putting `extensions` inside `external` directory, just to emphasise it's external and if we have more submodules in any future, it also end up in the same external directory.

jatinriverlane (approved):
Thanks @str-anger

PR #164, opened 2024-06-19
feat(api)!: API enchancement

samiralavi:
This PR is a for the renaming in the API to rename `alias` to `eid` and `id` to `uuid` for experiments.


jatinriverlane (approved):
Thanks @samiralavi

aqueductcore/frontend/src/components/organisms/ExperimentDetailsActionButtons/index.tsx
```diff
@@ -180,6 +180,7 @@ function ExperimentDetailsActionButtons({ isEditable, isDeletable, experimentDet
                         }
                         sx={{ minWidth: 140, ml: 1 }}
                         onClick={ToggleAddToFavourites}
+                        title="toggle-favourite"
```
str-anger, line 183: What is this? =)

extensions/python-example/solve_alpha.py
```diff
@@ -0,0 +1,63 @@
+import os
```
str-anger, line 1: Extensions are deleted and moved to `external/extensions` as a submodule

str-anger (changes requested):
A minor change requested -- you've added an extension file, probably accidentaly.

aqueductcore/frontend/src/components/organisms/ExperimentDetailsActionButtons/index.tsx
safeamiiir: It's a title which is used in tests

safeamiiir (approved):

extensions/python-example/solve_alpha.py
samiralavi: I have removed it.

str-anger (approved):

PR #170, opened 2024-06-26
fix(docker): Set the number of workers for release docker images to 4.

samiralavi:
This PR Set the number of workers for the release docker images to 4 to fix the issue with running extensions blocking the server thread from handling new requests.


jatinriverlane (approved):
Thanks @samiralavi

jatinriverlane:
I guess we would also need the increased number of workers in the `scripts/start_aqueduct_core.py` file.

str-anger (approved):

PR #171, opened 2024-06-26
feat(viewer): add html special type to preview

safeamiiir:
Make HTML files viewable in the GUI

### DIFF
--- a/aqueductcore/frontend/src/components/organisms/Attachments/Viewer/index.tsx
+++ b/aqueductcore/frontend/src/components/organisms/Attachments/Viewer/index.tsx
@@ -119,6 +119,14 @@ function Viewer({
               type: "JSON",
             });
           });
+        } else if (contentType?.includes("text/html")) {
+          response.text().then(() => {
+            console.log('fileURL', fileURL)
+            setInfo({
+              data: fileURL,
+              type: "HTML",
+            });
+          });
         } else if (contentType?.includes("text/x-markdown")) {
           response.text().then((data) => {
             setInfo({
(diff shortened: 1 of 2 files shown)


aqueductcore/frontend/src/components/organisms/Attachments/Viewer/index.tsx
```diff
@@ -175,6 +183,11 @@ function Viewer({
           <SpecialFilePreview>
             <ReactJson src={info.data} theme={theme.palette.mode === 'dark' ? 'bright' : 'bright:inverted'} />
           </SpecialFilePreview>
+          /* HTML */
+        ) : info?.type === "HTML" ? (
+          <SpecialFilePreview>
+            {info && <iframe style={{ height: 'inherit', width: "100%", border: 'none' }} src={String(info.data)} />}
```
jatinriverlane, line 189: I guess we can add `sandbox="allow-scripts"` for security purpose.

jatinriverlane (approved):
Thanks @safeamiiir

aqueductcore/frontend/src/components/organisms/Attachments/Viewer/index.tsx
safeamiiir: > Sandboxing is useless if the attacker can display content outside a sandboxed iframe — such as if the viewer opens the frame in a new tab. Such content should be also served from a separate origin to limit potential damage.

Well, just because of [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe) I find it not helpful at all, in their word, useless 😀

jatinriverlane: I agree, in our case, the origin will be same as well.

str-anger (approved):

PR #172, opened 2024-06-27
feat(plugins): Pass API token to the plugin execution environment if exists

samiralavi:
This PR enables passing API token to the plugin execution environment if exists.

### DIFF
--- a/aqueductcore/backend/models/extensions.py
+++ b/aqueductcore/backend/models/extensions.py
@@ -139,8 +139,9 @@ class ExtensionAction(BaseModel):
         my_env.update({key: str(val) for key, val in (extension.constants or {}).items()})
         my_env.update(params)
         my_env["aqueduct_url"] = extension.aqueduct_url
-        if extension.aqueduct_key is not None:
-            my_env["aqueduct_key"] = extension.aqueduct_key
+        if extension.aqueduct_api_token is not None:
+            my_env["API_TOKEN"] = extension.aqueduct_api_token
+
         cwd = Path.home()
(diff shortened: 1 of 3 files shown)


aqueductcore/backend/models/extensions.py
```diff
@@ -139,8 +139,9 @@ def execute(
         my_env.update({key: str(val) for key, val in (extension.constants or {}).items()})
         my_env.update(params)
         my_env["aqueduct_url"] = extension.aqueduct_url
-        if extension.aqueduct_key is not None:
-            my_env["aqueduct_key"] = extension.aqueduct_key
+        if extension.aqueduct_api_token is not None:
+            my_env["API_TOKEN"] = extension.aqueduct_api_token
```
str-anger, line 143: Is there a strong need in renaming the field?

```suggestion
            my_env["aqueduct_key"] = extension.aqueduct_api_token
```

str-anger (approved):
In general no problem with this, just if you rename the key, I will need to change the variable name in all extensions. Otherwise all good.

safeamiiir (approved):

aqueductcore/backend/models/extensions.py
samiralavi: the name of the environment variable that PyAqueduct picks up is `API_TOKEN`. So I used that instead of changing it to key. Also, technically API token and key are different concepts: https://www.gomomento.com/blog/api-keys-vs-tokens-whats-the-difference. What we do is API Token, not key.

PR #173, opened 2024-06-27
fix(compose): editing compose file and add setup hints

str-anger:
This PR edits an example compose file. Fixes some issues with extensions, adds volume to postgres, and gives some advises about setting the values.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -10,18 +10,20 @@ On this page we provide instructions for two ways to set up Aqueduct:
 ## Using Production Docker Images
 For the ease of deployment in production, Aqueduct is released as a production-ready container image. The docker image is available through Dockerhub, [`aqueducthub/aqueductcore`](https://hub.docker.com/r/aqueducthub/aqueductcore).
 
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -31,24 +33,45 @@ services:
       - /tmp/aqueduct_experiments:/tmp/aqueduct_experiments
       - type: bind
         # host folder for extensions
-        source: /aqueductcode/extensions
-        target: /workspace/extensions
+        source: /tmp/extensions
```
safeamiiir, line 36: I was just wondering if using `tmp` in the default config helps? what about just extensions?
```suggestion
        source: /extensions
```

safeamiiir (approved):
Helpful doc update

chididi (approved):

docs/setup.md
str-anger: This is the only folder where most probably you will by default have rights to create files. This strategy was chosen for files, so I just followed it.

PR #176, opened 2024-07-01
feat(upload): add upload button functionality

safeamiiir:
Uncomment upload button and let users upload files for experiments

### DIFF
--- a/aqueductcore/frontend/src/components/organisms/Attachments/Explorer/index.tsx
+++ b/aqueductcore/frontend/src/components/organisms/Attachments/Explorer/index.tsx
@@ -3,15 +3,16 @@ import InsertPhotoOutlinedIcon from "@mui/icons-material/InsertPhotoOutlined";
 import { Box, TableSortLabel, styled } from "@mui/material";
 import DataObjectIcon from "@mui/icons-material/DataObject";
 import TableContainer from "@mui/material/TableContainer";
-import { DragEvent, useMemo, useState } from "react";
 import TableBody from "@mui/material/TableBody";
 import TableCell from "@mui/material/TableCell";
 import TableHead from "@mui/material/TableHead";
 import TableRow from "@mui/material/TableRow";
+import { useMemo, useState } from "react";
 import Table from "@mui/material/Table";
 
 import { FileSelectContextType, SortOrder, selectedFileType } from "types/componentTypes";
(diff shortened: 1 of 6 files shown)


aqueductcore/frontend/src/components/organisms/Attachments/Explorer/index.tsx
```diff
@@ -112,48 +114,82 @@ function Explorer({
...
+          if (item.kind === "file") {
+            const file = item.getAsFile();
+            if (file) {
+              handleExperimentFileUpload(file)
+            }
+          }
+        });
+      } else {
+        [...ev.dataTransfer.files].forEach((file) => {
+          handleExperimentFileUpload(file)
```
str-anger, line 133: What this means is "I drop something, if some of these are files, I upload each of them separately", else "if I dropped a collection of files, upload them separately". How may this happen that dropped objects are in one of the collections?

aqueductcore/frontend/src/components/organisms/Attachments/index.tsx
```diff
@@ -15,54 +16,43 @@ font-size: 1.15rem;
...
   experimentFiles: ExperimentFileType[];
 }
 
 function Attachments({ experimentUuid, experimentFiles }: AttachmentProps) {
   const { selectedFile, setSelectedFile } = useContext(FileSelectStateContext)
-
+  const { handleExperimentFileUpload } = useFileUpload(experimentUuid)
+  function handleChangeFile(e: ChangeEvent<HTMLInputElement>) {
+    if (e.target.files) {
+      [...e.target.files].forEach(file => {
```
str-anger, line 29: Looks like a code duplication

str-anger (approved):
I cannot comment more =)

samiralavi (approved):
Thanks @safeamiiir, looks good, waiting to see it in action

aqueductcore/frontend/src/components/organisms/Attachments/index.tsx
safeamiiir: It's actually not.
This is when using `onChange` function in `<input />` 
and the one above is for `onDrop` function when doing Drag and Drop.

Why it's not? Because the `Event` in these two is different, we need to get the file and then use `handleUpload` function to upload them.

PR #177, opened 2024-07-03
fix(fetch): handle in case some error happened in Fetch

safeamiiir:
A quick fix when fetch failed due to the issues with the file

### DIFF
--- a/aqueductcore/frontend/src/hooks/useUploadFile.ts
+++ b/aqueductcore/frontend/src/hooks/useUploadFile.ts
@@ -53,6 +53,12 @@ function useFileUpload(experimentUuid: ExperimentDataType['uuid']) {
                         });
                 }
             })
+            // Somethings went wrong with the file
+            .catch((error) => {
+                toast.error(error.message ?? "Unable to upload file.", {
+                    id: "file_catch",
+                });
+            });
     }
     return { handleExperimentFileUpload }
 }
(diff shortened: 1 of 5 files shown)


jatinriverlane (approved):
Thanks @safeamiiir

aqueductcore/frontend/src/hooks/useUploadFile.ts
```diff
@@ -53,6 +53,12 @@ function useFileUpload(experimentUuid: ExperimentDataType['uuid']) {
                         });
                 }
             })
+            // Somethings went wrong with the file
+            .catch((error) => {
+                toast.error(error.message ?? "Unable to upload file.", {
```
jatinriverlane, line 58: The error message from the backend sends some relevant messages for some errors like Max Size Exceeded, Experiment Not Found, etc. I guess the error message can be shown here.

str-anger: May we have also the file name as well? Because I might drop a batch.

safeamiiir: file name added

safeamiiir: @jatinriverlane That error is also handled in line 44

str-anger (approved):

PR #186, opened 2024-07-09
feat(remove-files): add router view for removing files

jatinriverlane:
## Changes
* add router view for removing files from the experiment
* add relevant tests for remove experiment file

### DIFF
--- a/aqueductcore/backend/routers/files.py
+++ b/aqueductcore/backend/routers/files.py
@@ -214,12 +214,13 @@ async def remove_experiment_files(
     """Router for deleting file from an experiment"""
 
     request_body = await request.json()
-    file_list = set(request_body.get('file_list'))
+    file_list = request_body.get('file_list')
     if not file_list:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail="File list missing from request body",
         )
+    file_list = list(set(file_list))
 
     if file_list is None:
         raise HTTPException(
@@ -231,6 +232,7 @@ async def remove_experiment_files(
         for file_name in file_list:
             pathvalidate.validate_filename(file_name)
 
+        # check the experiment exists before deleting a file from it
         await get_experiment_by_uuid(
             user_info=context.user_info,
             db_session=context.db_session,
--- a/aqueductcore/backend/routers/files.py
+++ b/aqueductcore/backend/routers/files.py
@@ -204,7 +204,7 @@ async def upload_experiment_file(
     return JSONResponse({"result": f"Successfuly uploaded {file_name}"})
 
 
-@router.delete("/{experiment_uuid}/")
+@router.delete("/{experiment_uuid}")
 async def remove_experiment_file(
     request: Request,
     experiment_uuid: UUID,
@@ -212,15 +212,21 @@ async def remove_experiment_file(
 ) -> JSONResponse:
     """Router for deleting file from an experiment"""
 
-    file_names = request.headers.get("file_names")
+    file_names = request.headers.get("file_names").split(",")
     if file_names is None:
         raise HTTPException(
             status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail="Filenames header is missing."
         )
 
+    if len(file_names) == 0:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST, detail="Filenames header is empty."
+        )
+
     try:
         # check if experiment exists with the specified UUID, otherwise raises an exception.
-        pathvalidate.validate_filename(file_name) for file_name in file_names
+        for file_name in file_names:
+            pathvalidate.validate_filename(file_name)
 
         await get_experiment_by_uuid(
             user_info=context.user_info,
@@ -247,7 +253,7 @@ async def remove_experiment_file(
         if invalid_file_names:
             raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
-                detail=f"Files {invalid_file_names} as invalid",
+                detail=f"Files {','.join(invalid_file_names)} are invalid !"
             )
             
         for file_name in file_names:
@@ -260,4 +266,6 @@ async def remove_experiment_file(
             detail="Invalid file names.",
         ) from error
 
-    return JSONResponse({"result": f"Successfully removed {file_name}"})
+    return JSONResponse({
+        "result": f"Successfully removed {','.join(invalid_file_names)}"
+    })
--- a/aqueductcore/backend/routers/files.py
+++ b/aqueductcore/backend/routers/files.py
@@ -12,7 +12,7 @@ from streaming_form_data.targets import FileTarget
 from streaming_form_data.validators import MaxSizeValidator, ValidationError
 from typing_extensions import Annotated
 
-from aqueductcore.backend.context import ServerContext, context_dependency
+from aqueductcore.backend.context import ServerContext, FileList, context_dependency
 from aqueductcore.backend.errors import (
     AQDDBExperimentNonExisting,
     AQDMaxBodySizeException,
@@ -215,11 +215,11 @@ async def remove_experiment_files(
     request: Request,
     experiment_uuid: UUID,
     context: Annotated[ServerContext, Depends(context_dependency)],
+    file_list: FileList
 ) -> JSONResponse:
     """Router for deleting file from an experiment"""
 
-    request_body = await request.json()
-    file_list = request_body.get('file_list')
+    file_list = file_list.file_list
     if not file_list:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
@@ -229,8 +229,8 @@ async def remove_experiment_files(
 
     if file_list is None:
         raise HTTPException(
-            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
-            detail="File list is missing from the request body."
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="File list can not be empty."
         )
 
     try:
@@ -282,7 +282,7 @@ async def remove_experiment_files(
         else:
             raise HTTPException(
                 status_code=status.HTTP_400_BAD_REQUEST,
-                detail=f"File {file_name} is invalid !"
+                detail=f"File {file_name} is invalid!"
             )
 
     except AQDDBExperimentNonExisting as error:
--- a/aqueductcore/backend/routers/files.py
+++ b/aqueductcore/backend/routers/files.py
@@ -17,8 +17,9 @@ from aqueductcore.backend.errors import (
     AQDDBExperimentNonExisting,
     AQDMaxBodySizeException,
 )
-from aqueductcore.backend.services.utils import arr_humanize
+from aqueductcore.backend.services.utils import format_list_human_readable
 from aqueductcore.backend.services.constants import MARKDOWN_EXTENSIONS
+from aqueductcore.backend.context import UserScope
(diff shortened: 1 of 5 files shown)


aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +202,60 @@ async def upload_experiment_file(
...
+
+    if file_name is None:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail="Filename header is missing."
+        )
+
+    try:
+        pathvalidate.validate_filename(file_name)
+
+        await get_experiment_by_uuid(
```
str-anger, line 225: Is this not accepting a result for a reason?

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +202,60 @@ async def upload_experiment_file(
         ) from error
 
     return JSONResponse({"result": f"Successfuly uploaded {file_name}"})
+
+
+@router.delete("/{experiment_uuid}")
+async def remove_experiment_file(
```
str-anger, line 208: May I ask you to separate router function and actual business logic (file deletion). It will probably sit nice in `Experiment.remove_file(filename)` method.

str-anger (changes requested):

aqueductcore/backend/routers/files.py
jatinriverlane: It's not accepting any result, this is just a validation. It checks if experiment with the given UUID is present or not. If it does not exist, it raises an exception from Line 270.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +203,79 @@ async def upload_experiment_file(
...
+    if file_list is None:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="File list is missing from the request body."
+        )
+
+    try:
+        for file_name in file_list:
+            pathvalidate.validate_filename(file_name)
+
```
str-anger, line 233:
```suggestion

# check the experiment exists before deleting a file from it
```

aqueductcore/backend/services/utils.py
```diff
@@ -73,3 +73,14 @@ def is_tag_valid(tag: str) -> bool:
     regex = recompile(pattern)
 
     return bool(regex.match(tag))
+
+
+def arr_humanize(arr: list[str | int]) -> str:
```
str-anger, line 78: test on this, please?

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +203,79 @@ async def upload_experiment_file(
         ) from error
 
     return JSONResponse({"result": f"Successfuly uploaded {file_name}"})
+
+
+@router.delete("/{experiment_uuid}")
+async def remove_experiment_files(
```
str-anger, line 209: I will copy my comments from the previous review, as it was not addressed.

May I ask you to separate router function and actual business logic (file deletion). It will probably sit nice in `Experiment.remove_file(filename)` method?

str-anger: And then - a test please.

str-anger (changes requested):

aqueductcore/backend/services/utils.py
jatinriverlane: Have added test for this

aqueductcore/backend/routers/files.py
jatinriverlane: The structure of the file based features are monolithic functions. The upload_file and download_file as also in the same pattern. For refactoring, we can divide the functions in this file to multiple smaller blocks. I guess it can be picked in a different PR.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +203,81 @@ async def upload_experiment_file(
...
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail="The specified experiment was not found.",
+        ) from error
+    except pathvalidate.ValidationError as error:
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="Invalid file names.",
+        ) from error
+
+    return JSONResponse({"result": f"Successfully deleted {arr_humanize(file_list)}"})
```
samiralavi, line 283: Does this really need its own utility function if it is only used once? Also, the function name is not clear.

samiralavi (approved):
Thanks @jatinriverlane, the logic is fine. Could you please re-evaluate to implement the functionality as a GraphQL mutation instead of a new REST API? As we discussed, we only use REST API only for file upload and download as an exception as the GraphQL library didn't expose file streaming. If it is something that can be done quickly, I suggest doing it now, otherwise, we can later refactor it.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +203,81 @@ async def upload_experiment_file(
         ) from error
 
     return JSONResponse({"result": f"Successfuly uploaded {file_name}"})
+
+
+@router.delete("/{experiment_uuid}")
```
samiralavi, line 208: If you want to stick to REST contract, I suggest using DELETE operation with `"/{experiment_uuid}/{file_name}"` for single file deletion. If you want multiple files deletion in one call, change it to POST, as it is not a common practice to add data to the request body with DELETE verb.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +203,81 @@ async def upload_experiment_file(
...
+async def remove_experiment_files(
+    request: Request,
+    experiment_uuid: UUID,
+    context: Annotated[ServerContext, Depends(context_dependency)],
+) -> JSONResponse:
+    """Router for deleting file from an experiment"""
+
+    request_body = await request.json()
+    file_list = request_body.get('file_list')
+    if not file_list:
```
samiralavi, line 218: Please make sure this operation is secured by a specific user scope and the user scope should be checked for that.

samiralavi (changes requested):
Thanks @jatinriverlane. I have requested changes to the API contract and how it should be called. Also, a new user scope for file deletion is needed as this is a critical operation that should be secured.

aqueductcore/backend/routers/files.py
jatinriverlane: I agree, as we wish to support multiple file deletion in a single request, this is updated to being a POST request with the file list in the request body.

aqueductcore/backend/routers/files.py
jatinriverlane: This is used at 2 places. However, the name is not appropriate. The name of this utility function is now converted to `format_list_human_readable`

aqueductcore/backend/routers/files.py
jatinriverlane: Sure @samiralavi , have added user scope check

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+# pylint: disable=too-many-return-statements,too-many-branches
+@router.post("/{experiment_uuid}/delete_files")
+async def remove_experiment_files(
+    request: Request,
+    experiment_uuid: UUID,
+    context: Annotated[ServerContext, Depends(context_dependency)],
+) -> JSONResponse:
+    """Router for deleting file from an experiment"""
+
+    request_body = await request.json()
```
samiralavi, line 221: FastAPI already provides a safe way to get arguments with type validation. Please rewrite that using that functionality:
https://fastapi.tiangolo.com/tutorial/body-fields/

samiralavi (changes requested):
Thanks @jatinriverlane. I am happy with the changes except the way you are handling request body. I have put the a comment to refactor using FastAPI models.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="File list missing from request body",
+        )
+    file_list = list(set(file_list))
+
+    if file_list is None:
+        raise HTTPException(
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+            detail="File list is missing from the request body."
```
safeamiiir, line 233: I was wondering maybe we should have a different error message compared to `400`
```suggestion
            detail="File list can not be empty"
```

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+            user_info=context.user_info,
+            db_session=context.db_session,
+            experiment_uuid=experiment_uuid,
+        )
+
+        if UserScope.EXPERIMENT_DELETE_ALL not in context.user_info.scopes:
+            if UserScope.EXPERIMENT_DELETE_OWN not in context.user_info.scopes:
+                raise AQDPermission(
+                    "The user doesn't have permission to delete files from experiment"
+                )
```
safeamiiir, line 250: Don't you agree we can do this logic in a function?
something like what we have in the frontend `isUserAbleToDeleteExperiment`

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+                detail=f"File(s) not found - {format_list_human_readable(invalid_files)}"
+            )
+
+        dest_file_path = os.path.join(experiment_dir, file_name)
+        if os.path.exists(dest_file_path) and os.path.isfile(dest_file_path):
+            os.remove(dest_file_path)
+        else:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"File {file_name} is invalid !"
```
safeamiiir, line 285:
```suggestion
                detail=f"File {file_name} is invalid!"
```

safeamiiir (approved):

aqueductcore/backend/routers/files.py
jatinriverlane: @safeamiiir This block is used only once in Aqueduct backend. I guess it'll better to convert it to a function once we have more than one usage of such block.

aqueductcore/backend/routers/files.py
str-anger: I don't feel like "other functions are already monolithic" is a valid excuse for not stopping doing this =)

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+@router.post("/{experiment_uuid}/delete_files")
+async def remove_experiment_files(
+    request: Request,
+    experiment_uuid: UUID,
+    context: Annotated[ServerContext, Depends(context_dependency)],
+    file_list: FileList
+) -> JSONResponse:
+    """Router for deleting file from an experiment"""
+
+    file_list = file_list.file_list
```
str-anger, line 222: 1. I see duplication with 189....

str-anger (changes requested):

aqueductcore/backend/routers/files.py
jatinriverlane: @str-anger Actually, feat/remove-experiment-file-frontend was branched off from this branch only. Because of that, the changes are visible in both PRs.

aqueductcore/backend/routers/files.py
jatinriverlane: As discussed, this will be picked in a different PR...

samiralavi (approved):
Thanks @jatinriverlane for the changes.

PR #187, opened 2024-07-09
fix(graphql): fix keyFields of GraphQL

safeamiiir:
- Fix keyFields after renaming `id`


str-anger (approved):
No idea what's going on, but looks legit =)

safeamiiir:
> No idea what's going on, but looks legit =)

It has to be a unique field to enable [Apollo GraphQL](https://www.apollographql.com/docs/react/caching/cache-configuration/#customizing-cache-ids) manages caching of the app.
Since we had a renaming some of this has to be fixed if there's a problem with current behaviour.

jatinriverlane (approved):
Thanks @safeamiiir

PR #189, opened 2024-07-14
feat(delete-file-frontend): Functionality to delete files from experiment

jatinriverlane:
## Changes
* add delete file function
* add delete file from experiment button
* add relevant tests for delete file functionality

This branch was created from https://github.com/AqueductHub/aqueductcore/pull/186


aqueductcore/backend/context.py
```diff
@@ -35,6 +35,12 @@ class UserInfo(BaseModel):
     token: Optional[str] = None
 
 
+class FileList(BaseModel):
```
str-anger, line 38: What stops us from using `List[str]` instead? This wrapper brings no additional capabilities.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
     return JSONResponse({"result": f"Successfuly uploaded {file_name}"})
+
+
+# pylint: disable=too-many-return-statements,too-many-branches
+@router.post("/{experiment_uuid}/delete_files")
+async def remove_experiment_files(
+    request: Request,
+    experiment_uuid: UUID,
+    context: Annotated[ServerContext, Depends(context_dependency)],
+    file_list: FileList
```
str-anger, line 218:
```suggestion
    file_list: List[str]
```

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+) -> JSONResponse:
+    """Router for deleting file from an experiment"""
+
+    file_list = file_list.file_list
```
str-anger, line 222: ... Especially if you have this line. This changes type of the file_list object (was FileList, became List[str])

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+
+    if not file_list:
```
str-anger, line 224: This line also triggers for `None`.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="File list missing from request body",
+        )
+
+    if file_list is None:
```
str-anger, line 230: This is impossible, if you did not make the list optional.

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+
+        if UserScope.EXPERIMENT_DELETE_ALL not in context.user_info.scopes:
+            if UserScope.EXPERIMENT_DELETE_OWN not in context.user_info.scopes:
+                raise AQDPermission(
+                    "The user doesn't have permission to delete files from experiment"
+                )
+
+            if experiment.created_by != context.user_info.uuid:
+                raise AQDDBExperimentNonExisting(
+                    "The user doesn't have permission to delete files from this experiment."
```
str-anger, line 254: Should we add user name as well to the message?

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+            if not os.path.exists(abs_file_path) or not os.path.isfile(abs_file_path):
+                invalid_files.append(file_name)
+
+        if invalid_files:
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail=f"File(s) not found - {format_list_human_readable(invalid_files)}"
+            )
+
+        dest_file_path = os.path.join(experiment_dir, file_name)
```
str-anger, line 279: Is this outside the loop? files name is the variable name from the loop on line 268
Looks like `for` is missing

aqueductcore/backend/routers/files.py
```diff
@@ -202,3 +207,93 @@ async def upload_experiment_file(
...
+        if os.path.exists(dest_file_path) and os.path.isfile(dest_file_path):
+            os.remove(dest_file_path)
```
str-anger, line 281: Am I right this is a second-time validation for the case of parallel competing process doing the same? I'd probably let it fail without re-check.

tests/unittests/test_files.py
```diff
@@ -422,3 +422,162 @@ async def override_context_dependency() -> AsyncGenerator[ServerContext, None]:
...
+
+    non_existing_file = "non_existing_file.zip"
+    request_body = {"file_list": [non_existing_file]}
+
+    response = client.post(
+        f"{settings.api_prefix}{settings.files_route_prefix}/{str(db_experiment.uuid)}/delete_files",
+        json=request_body,
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
+    assert response.json() == {"detail": f"File(s) not found - {non_existing_file}"}
```
str-anger, line 583: 1. May you please add/edit tests to capture all exceptional situations?
2. Also, probably handle
  a. longer-than 1 list
  b. partial matches
  c. a folder instead of the file

str-anger (changes requested):
Some changes may probably make it better!

aqueductcore/backend/context.py
jatinriverlane: @str-anger https://fastapi.tiangolo.com/tutorial/body-fields/ mentions the recommended way of parsing request body as a pydantic model. Hence, we are using this.

aqueductcore/backend/routers/files.py
jatinriverlane: @str-anger An alternate change for this would be to set the name of parameter to be something similar to request_body.

aqueductcore/backend/routers/files.py
jatinriverlane: Having request body as parameter name would make it favourable to use
```python
file_list = request_body.file_list
```

aqueductcore/backend/routers/files.py
jatinriverlane: So, if `file_list` is None, pydantic will return error by itself, with status code 422.

i.e. in case `file_list` is `None`, code will not reach this point. This is handling just the case for empty `file_list` array.

aqueductcore/backend/routers/files.py
jatinriverlane: Right @str-anger , I'll remove this block.

aqueductcore/backend/routers/files.py
jatinriverlane: In other APIs, we haven't mentioned the username in the error message. Having such error just makes it similar to other responses. For eg.
* The user doesn't have the permission to remove the experiment.
* The user doesn't have the required permission(s) to create experiments.

However, it's easy to include username in error message. Do you suggest that we should add username ?

aqueductcore/backend/routers/files.py
jatinriverlane: Right, there's no need to validate during the delete operation. Just the initial validation is fine.

aqueductcore/backend/routers/files.py
jatinriverlane: 🤦‍♂️ Slipped my mind. Have fixed this now in the backend PR https://github.com/AqueductHub/aqueductcore/pull/186

tests/unittests/test_files.py
jatinriverlane: Sure, I've implemented `longer-than 1 list` and `partial matches` tests. I'm thinking about other tests that can be created for this. 

But, I don't see `folder instead of the file` happening in real life.

aqueductcore/backend/context.py
samiralavi: @jatinriverlane does this need to be in the `context.py` module?

samiralavi (approved):
I have already reviewed and approved the backend changes in the previous PR. I approve here the frontend changes, but please make sure @safeamiiir reviews this as he has the the FE knowledge the same as you.

aqueductcore/backend/context.py
jatinriverlane: @samiralavi As discussed, I've moved it to `backend/routers/files.py`

aqueductcore/frontend/src/components/organisms/Attachments/index.tsx
```diff
@@ -49,19 +63,23 @@ function Attachments({ experimentUuid, experimentFiles }: AttachmentProps) {
...
           >
             Delete
           </BorderedButtonWithIcon>
-        </Grid> */}
+          <DeleteExperimentFileModal
+            open={isDeleteExperimentFileModalOpen}
+            onClose={handleCloseDeleteExperimentFileModal}
+            handleDeleteExperimentFile={handleFileDelete}
+          />
+        </Grid>}
```
safeamiiir, line 82: What do you think about keeping the Divider and disabling the delete button?

```jsx
        <Grid item>
          <Divider orientation="vertical" />
        </Grid>
        <Grid item>
          <BorderedButtonWithIcon
            variant="outlined"
            size="small"
            color="neutral"
            sx={{ ml: 0.5 }}
            onClick={handleOpenDeleteExperimentFileModal}
            startIcon={<DeleteOutlineOutlinedIcon />}
            disabled={!selectedFile}
          >
            Delete
          </BorderedButtonWithIcon>
          <DeleteExperimentFileModal
            open={isDeleteExperimentFileModalOpen}
            onClose={handleCloseDeleteExperimentFileModal}
            handleDeleteExperimentFile={handleFileDelete}
          />
        </Grid>
```

safeamiiir (approved):
Thank you Jatin!
I just added a suggestion on delete button behaviour.

aqueductcore/frontend/src/components/organisms/Attachments/index.tsx
jatinriverlane: 🤔 The current behaviour seems to be fine. Showing a disabled button when no file is selected might be confusing. A disabled button might show that the button is not functional.

safeamiiir (approved):
Thanks!

tests/unittests/test_files.py
```diff
@@ -466,6 +467,104 @@ async def override_context_dependency() -> AsyncGenerator[ServerContext, None]:
...
+        file_path = os.path.join(experiment_dir, experiment_file_name)
+        with open(file_path, "wb") as file_writer:
+            file_writer.write(os.urandom(settings.upload_max_file_size_KB * BYTES_IN_KB))
+
+    request_body = {"file_list": [*experiment_file_names, invalid_file]}
+    response = client.post(
+        f"{settings.api_prefix}{settings.files_route_prefix}/{str(db_experiment.uuid)}/delete_files",
+        json=request_body,
+    )
+    assert response.status_code == status.HTTP_400_BAD_REQUEST
```
str-anger, line 564: I'd say that 400 is generic and correct, but may be confused with just a messy input. What about 409, 417? https://umbraco.com/knowledge-base/http-status-codes/

jatinriverlane: @str-anger These status codes seems to be related to request headers and server states. 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

400 seems to be the better option. You can read from the above link, it has more explanation for this.

str-anger: 409 looks reasonable: 
> [409 Conflict](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409)
This response is sent when a request conflicts with the current state of the server.

jatinriverlane: Sure, 409 seems to be the right status code for this case. Thanks @str-anger

aqueductcore/frontend/src/components/organisms/Attachments/index.tsx
jatinriverlane: @safeamiiir As Tom suggested, I've implemented your suggestion to enable /disable Delete button based on file selection

PR #193, opened 2024-07-17
feat(private-extensions): we need to be able to get private extensions easily

str-anger:
This is how you would download extensions as folders in github:

```
 ./scripts/add_extensions_from.sh git@github.com:AqueductHub/aqueduct-extensions-private.git main
```


tomhartley (approved):
Looks good to me!

samiralavi (changes requested):
Could you please add the following items to the script, similar to other scripts:
- shebang
- comments for what the scripts does and how it should be used
- if you are assuming the script should run in the root of the project, make sure this happens disregarding the caller current working directory. You can look at other scripts to see how we have done it.

samiralavi (approved):
Thanks @str-anger for the improvement. I assume this script works only in the dev environment.

PR #196, opened 2024-07-25
fix(permission): fix user permission to remove file

safeamiiir:
Check if the user has access to the experiment via calling the function `get_experiment_by_uuid`.


jatinriverlane (approved):
Thanks for fixing this Amirreza

samiralavi (approved):
Thanks @safeamiiir it looks good

PR #197, opened 2024-07-29
feat(jobs): using celery executors for extensions

str-anger:


aqueductcore/backend/services/extensions_executor.py
```diff
@@ -21,6 +20,7 @@
     build_experiment_dir_absolute_path,
     get_experiment_by_eid,
 )
+from aqueductcore.backend.services.task_executor import execute_blocking, execute_non_blocking
```
samiralavi, line 23: I couldn't see any use of `execute_non_blocking` function

str-anger:
This PR is a reference for the new Jobs functionality. Won't be merged.

PR #200, opened 2024-08-06
feat(celery)!: Support for Celery tasks

samiralavi:
This PR adds the celery and flower dependencies to Aqueduct Core with a basic module to initialise the Celery app, `task_executor.py`. New settings variables are added for RabbitMQ configuration via environment variables. Aqueduct CLI is also extended to support two new commands: 1) Running Celery worker, and 2) Running Flower UI. As these commands are bundled into the Aqueduct python package that gets installed in the release docker image of Aqueduct, the users using the same docker image, can spin three containers: 1) Aqueduct Core, 2) Celery Worker, and 3) Flower UI.  

The Celergy backend is also configured to persist tasks into Postgres database. For now, I have chosen the same database as the experiments are stored in, which simplifies establishing foreign keys relation ship between experiments and tasks. Also, it simplifies database migration as the user has to migrate only a single database between different versions.


aqueductcore/backend/services/task_executor.py
```diff
@@ -0,0 +1,13 @@
...
+from celery import Celery
+
+from aqueductcore.backend.settings import settings
+
+celery_app = Celery(
+    "tasks",
+    broker=f"amqp://{settings.rabbitmq_username}:{settings.rabbitmq_password}"
+    "@{settings.rabbitmq_host}:{settings.rabbitmq_port}",
+    backend=f"db+postgresql://{settings.postgres_username}:{settings.postgres_password}"
+    "@{settings.postgres_host}:{settings.postgres_port}/{settings.postgres_db}",
```
str-anger, line 12: Being honest, I feel this like a narrowing of configurability. E.g. you "nail" "`db+postgresql:`" and "`amqp:`" protocols. Also, instead ob having 2 parameters, you have 9. This is not a huge problems, but my preference is to have 2 connection strings configurable by user.

```suggestion
    broker=settings.celery_message_queue,
    backend=settings.celery_backend,
```

aqueductcore/backend/settings.py
```diff
@@ -35,6 +35,17 @@ class Settings(BaseSettings):
...
+    """Rabbitmq hostname."""
+    rabbitmq_port: int
+    """Rabbitmq port."""
+    rabbitmq_management_port: Optional[int] = None
+    """Rabbitmq port."""
+    rabbitmq_username: str
+    """Rabbitmq username."""
+    rabbitmq_password: str
+    """Rabbitmq password."""
+
```
str-anger, line 48:
```suggestion
    celery_message_queue: str
    """Message queue connection string."""
    celery_backend: str
    """Database connection string."""
    
```

aqueductcore/cli/main.py
```diff
@@ -88,3 +89,40 @@ def importer(
...
+        [
+            "celery",
+            "-A",
+            "aqueductcore.backend.services.task_executor.celery_app",
+            "worker",
+            "-l",
+            "info",
+        ],
+    ) as process:
+        process.wait()
```
str-anger, line 107: Out of curiosity - do we have to wait on the process? I think this has something to do with docker container status.

aqueductcore/cli/main.py
```diff
@@ -88,3 +89,40 @@ def importer(
...
+        raise typer.Exit(code=1)
+
+    with subprocess.Popen(
+        [
+            "celery",
+            "-A",
+            "aqueductcore.backend.services.task_executor.celery_app",
+            "flower",
+            f"--broker-api=http://{settings.rabbitmq_username}:{settings.rabbitmq_password}"
+            f"@{settings.rabbitmq_host}:{settings.rabbitmq_management_port}/api/vhost",
```
str-anger, line 125: May this also be configured as a single connection string?

containers/dev/docker-compose.yaml
```diff
@@ -30,3 +31,9 @@ services:
       - POSTGRES_DB=aqueductcore_dev
     expose:
       - 5432
+
+  rabbitmq:
+    image: rabbitmq:3-management
+    ports:
+      - "5672:5672"
+      - "15672:15672"
```
str-anger, line 39: Should we configure this container and pass arbitrary usernames and passwords? Is there the case that these ports are exposed to end users?

aqueductcore/cli/main.py
samiralavi: By default `subprocess.Popen` call the subprocess in non-blocking mode but here we want to intentionally tie main process to the subprocess by waiting on it.

aqueductcore/cli/main.py
samiralavi: I agree, connection strings are also a good alternative. I am going to update it as I don't have a hard opinion on it, both ways are good.

aqueductcore/backend/services/task_executor.py
samiralavi: I have added the connection string for the message queue, but for the backend, the already existing Postgres environment variables were suffice, so I kept them as before. Hopefully in future, we change the Postgres connection config into a connection string as well.

jatinriverlane (approved):
Seems fine, great work @samiralavi

aqueductcore/backend/services/task_executor.py
str-anger: What stops you now? =)

aqueductcore/backend/services/task_executor.py
```diff
@@ -0,0 +1,12 @@
+"""Celery task execution."""
+
+from celery import Celery
+
+from aqueductcore.backend.settings import settings
+
+celery_app = Celery(
+    "tasks",
+    broker=settings.celery_message_queue,
+    backend=f"db+postgresql://{settings.postgres_username}:{settings.postgres_password}"
```
str-anger, line 10: It looks like a very strange compromise. What I may suggest, is to have a:

```
class Settings:
...
    @property
    def celery_backend(self):
        return (
            f"db+postgresql://{self.postgres_username}:{self.postgres_password}"
            f"@{self.postgres_host}:{self.postgres_port}/{self.postgres_db}"
        )
```

which will simplify a future replacement with a parameter ;-)

str-anger (approved):
A minor request, which is non blocking any way.

aqueductcore/backend/services/task_executor.py
samiralavi: Nice suggestion, it is done.

PR #201, opened 2024-08-06
fix(docker): Update reference release docker compose

samiralavi:
This PR updates the reference docker-compose to add celery worker and flower alongside rabbitmq. Traefik reverse proxy is also added to let the user access different services via the host address `*.localhost`. Traefik automatically handles this DNS addition to the host, which I found very interesting! We might later change it but it is a good starting point.

### DIFF
--- a/scripts/release/docker-compose.yaml
+++ b/scripts/release/docker-compose.yaml
@@ -1,26 +1,58 @@
-version: '3'
+version: "3"
+
+x-aqueduct-common: &aqueduct-common
+  platform: linux/amd64
+  image: aqueducthub/aqueductcore-dev:latest
+  restart: always
+  depends_on:
+    - postgres
+    - rabbitmq
+  environment:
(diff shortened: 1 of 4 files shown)


jatinriverlane (approved):
Trafiek seems to be a very good solution for reverse proxy for our system, thanks Amir

scripts/release/docker-compose.yaml
```diff
@@ -1,26 +1,58 @@
-version: '3'
+version: "3"
+
+x-aqueduct-common: &aqueduct-common
+  platform: linux/amd64
+  image: aqueducthub/aqueductcore-dev:latest
+  restart: always
```
str-anger, line 6: Does this mean we expect all 3 child containers to automatically restart?

str-anger (approved):
I mean, I only understand it partially. Looks ok =)

scripts/release/docker-compose.yaml
samiralavi: Exactly!

PR #202, opened 2024-08-07
feat(celery-jobs): submit jobs to celery

str-anger:


aqueductcore/backend/models/extensions.py
```diff
@@ -148,22 +153,16 @@ def execute(
...
-            code = proc.returncode
-            return ExtensionExecutionResult(
-                return_code=code,
-                stdout=out.decode(),
-                stderr=err.decode(),
-            )
+            rel_python = Path(python).relative_to(cwd)
+            rich_script = rich_script.replace("$python ", f"{rel_python} ")
+
+        task = execute_task(
```
str-anger, line 159: This is the real core change: from processes to tasks

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -111,7 +117,7 @@ async def execute_extension(
         extension: str,
         action: str,
         params: List[List[str]],
-    ) -> ExtensionExecutionResult:
+    ) -> TaskInfo:
```
str-anger, line 120: Now the new, richer class is returned.

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -134,22 +140,51 @@ async def execute_extension(
...
-
         result = ExtensionsExecutor.execute(extension, action, dict_params)
 
-        await ExtensionsExecutor.save_log_to_experiment(
-            context=context,
-            eid=eid,
-            log_filename=log_filename,
-            result=result,
+        ### data population after submission
+        experiment = await get_experiment(
```
str-anger, line 146: Here and below is the code to populate the new class `TaskInfo`. As @safeamiiir asked, it comes together with experiment field.

aqueductcore/backend/services/extensions_executor.py
```diff
@@ -179,34 +176,3 @@ def execute(cls, extension: str, action: str, params: dict) -> ExtensionExecutio
             params=params,
             python=python,
         )
-
-    @classmethod
-    async def save_log_to_experiment(
```
str-anger, line 184: We don't save log any more.

aqueductcore/backend/services/task_executor.py
```diff
@@ -1,11 +1,120 @@
...
+        out, err = proc.communicate(timeout=None)
+        code = proc.returncode
+    return (
+        code,
+        err.decode("utf-8"),
+        out.decode("utf-8"),
+    )
+
+
+def execute_task(
```
str-anger, line 87: This is the code to execute tasks. Be careful. if any changes are made inside this class, there is a high chance you will need to rebuild a dev docker environment: as this file is used in both aqueduct AND celery worker.

samiralavi: There is no need to do that in the dev environment. It picks up the fresh code as the Python package is installed in edit mode via Poetry. All you need is to restart the worker.

tests/integration/test_extensions_executor.py
```diff
@@ -0,0 +1,57 @@
+import shutil
+
+import pytest
+from aqueductcore.backend.services.extensions_executor import (
+    ExtensionsExecutor, VENV_FOLDER, PYTHON_BINARY)
+
+
+class TestExtensionExecutor:
+
+    def test_extension_echo(self):
```
samiralavi, line 10: I guess these tests pass until the job gets submitted to the job queue, but there won't be a worker to pick it up.

samiralavi (approved):
Great work @str-anger, only a few minor comments.

safeamiiir (commented):
I'm happy with the API and as a discussed the std_our and std_err are no longer needed and could be removed in the future.
I'd rather Jatin gives the second approval as he could give more valuable comments.

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -134,22 +140,51 @@ async def execute_extension(
...
+            task_id=result.task_id,
+            experiment=experiment,
+            username=context.user_info.username,
+            extension_name=extension,
+            action_name=action,
+            parameters=parameters,
+            task_status=TaskStatus(result.status),
+
+            receive_time=result.receive_time,
+            started_time=None,
```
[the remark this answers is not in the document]
str-anger: I think yes, because obviously start time may only be assigned what status changes to STARTED =)

tests/integration/test_extensions_executor.py
str-anger: Yes. We only check that the job is submitted now. I will probably extend it up to retrieving a result.

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -134,22 +142,51 @@ async def execute_extension(
...
         )
-        return ExtensionExecutionResult(
-            return_code=result.return_code,
-            stdout=result.stdout,
-            stderr=result.stderr,
-            log_experiment=eid,
-            log_file=log_filename,
+        extension_info = ExtensionInfo.from_extension(extension_object)
+        action_info = None
+        parameters = []
```
jatinriverlane, line 157: I guess this line is not required and can be removed

aqueductcore/backend/services/task_executor.py
```diff
@@ -1,11 +1,130 @@
...
+WAITING_TIME = 2
+
 celery_app = Celery(
     "tasks",
     broker=settings.celery_message_queue,
     backend=settings.celery_backend,
 )
+
+
+@dataclass
```
jatinriverlane, line 26: Just a suggestion - I guess we can just not use data class here, mainly because it won't be any different. We are not using dataclasses in Aqueduct and here we aren't using any feature of data class.

jatinriverlane (approved):
Great job Stas, thanks

aqueductcore/backend/services/task_executor.py
```diff
@@ -1,11 +1,130 @@
...
+def execute_task(
+    extension_directory_name: str,
+    shell_script: str,
+    execute_blocking: bool = False,
+    **kwargs
+) -> TaskProcessExecutionResult:
+    """Execute a task and wait until finished"""
+
+    receive_time = datetime.now()
+    task = run_executable.delay(
```
jatinriverlane, line 123: @str-anger I have a question about the values to be passed with the task. For eg. there's an optional parameter `retries` that can be passed with delay. Is there a way to pass such parameters ? Can we use `**kwargs` to handle those values ?

aqueductcore/backend/services/task_executor.py
str-anger: I think a constructor of field names is used, but I will reconsider this.

aqueductcore/backend/services/task_executor.py
str-anger: 1. No we can/could not add this via kwargs, as `.delay()` method is a shortcut implementation which does not allow execution parameters.
2. I've switched to explicit method call `.apply_async()` which allows to do this. But I see no reasonable parameter. Also, we expect these containers run next to each other, and not being able to schedule a job is kind of a strong issues, which most likely won't be resolved with a retry (at the same host).
3. What I did -- I've added a comment, how this may be changed if there is a need of any policy in future.

PR #203, opened 2024-08-08
feat(task-cancel): task cancellation endpoint

str-anger:
This PR adds and endpoint for task cancellation. A few things to note:
1. According to functional specification, we should expect KeyboardInterrupt exception (SIGINT), but this this signal task is not cancelled: it changes status to REVOKED, and then, after process is finished,  becomes successful. To actually kill the job SIGTERM and SIGKILL are suitable.
2. There are a few comments left, that after implementation of persistence, some code should be added to populate `TaskInfo` objects.
3. TaskExecutor and mutation call tests added.

### DIFF
--- a/aqueductcore/backend/routers/graphql/mutations_schema.py
+++ b/aqueductcore/backend/routers/graphql/mutations_schema.py
@@ -191,3 +191,11 @@ class Mutation:
             # TODO: remove after frontend change, obsolete field
             return_code=0,
         )
+
+    @strawberry.mutation
+    async def cancel_task(
+        self,
+        info: Info,
+        task_id: UUID,
+    ) -> TaskInfo:
+        raise NotImplementedError()
--- a/aqueductcore/backend/routers/graphql/mutations_schema.py
+++ b/aqueductcore/backend/routers/graphql/mutations_schema.py
@@ -35,6 +35,7 @@ from aqueductcore.backend.routers.graphql.types import (
     TaskInfo,
 )
 from aqueductcore.backend.services.extensions_executor import ExtensionsExecutor
+from aqueductcore.backend.services.task_executor import revoke_task
 from aqueductcore.backend.routers.graphql.resolvers.experiment_resolver import (
     get_experiment,
 )
@@ -134,6 +135,7 @@ class Mutation:
         """
         dict_params = dict(params)
         context = cast(ServerContext, info.context)
+        username = "" if context.user_info is None else context.user_info.username
         extension_object = ExtensionsExecutor.get_extension(extension)
         extension_object.aqueduct_api_token = context.user_info.token
 
@@ -173,7 +175,7 @@ class Mutation:
         return TaskInfo(
             task_id=result.task_id,
             experiment=experiment,
-            username=context.user_info.username,
+            username=username,
             extension_name=extension,
             action_name=action,
             parameters=parameters,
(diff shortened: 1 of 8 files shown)


aqueductcore/backend/services/task_executor.py
```diff
@@ -126,5 +131,18 @@ def execute_task(
...
+
+def revoke_task(task_id: str, terminate: bool) -> TaskProcessExecutionResult:
+    """Cancel the task and update the status."""
+
+    task = AsyncResult(task_id)
+    # note: SIGINT does not lead to task abort. Id you send
+    # KeyboardInterupt (SIGINT), it will not stop, and the
+    # exception does not propagate.
+    task.revoke(terminate=terminate, signal="SIGTERM")
+    time.sleep(WAITING_TIME)
```
samiralavi, line 147: Please change the sleep to be async, so it doesn't block the web server from processing other tasks:
```python
import asyncio
await asyncio.sleep(WAITING_TIME)
```

samiralavi: Anyhow, I think this wait can be avoided.

samiralavi (approved):
Thanks @str-anger, I approved this PR, but please consider my proposal to use async sleep instead.

aqueductcore/backend/routers/graphql/mutations_schema.py
```diff
@@ -191,3 +193,51 @@ async def execute_extension(
...
+
+        Args:
+            info: context, which includes database connection.
+            task_id: ID of the task to cancel.
+
+        Returns:
+            TaskInfo: status of the task after cancellation.
+        """
+        context = cast(ServerContext, info.context)
+        task_state = revoke_task(task_id=str(task_id), terminate=True)
```
jatinriverlane, line 213: I guess we can rethink about the `terminate` parameter here. There might be some extensions that would want the terminate value to be False. For instance, there might be some task that sends a request for a long running job and cancelling the job abruptly would result in some large intermediate files in disk that would need to be cleaned... 

🧐

jatinriverlane (approved):
Thanks @str-anger , great work...

aqueductcore/backend/services/task_executor.py
```diff
@@ -126,5 +131,18 @@ def execute_task(
...
+
+
+def revoke_task(task_id: str, terminate: bool) -> TaskProcessExecutionResult:
+    """Cancel the task and update the status."""
+
+    task = AsyncResult(task_id)
+    # note: SIGINT does not lead to task abort. Id you send
+    # KeyboardInterupt (SIGINT), it will not stop, and the
+    # exception does not propagate.
+    task.revoke(terminate=terminate, signal="SIGTERM")
```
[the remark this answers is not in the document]
str-anger: Thank you for the comment. Indeed, there was expectation to work with flower. There are two tasks, which may resolve this difference (TT-107, TT-123). If these tasks will benefit from switching to Flower, I think this may be done. Currently code has no infrastructure to talk to flower, it is yet to appear.

aqueductcore/backend/routers/graphql/mutations_schema.py
str-anger: I will add this comment to TT-102. Your comment is 100% valid, and I think that implementation of that ticket is right place and time to consider this.

safeamiiir:
Like to see these async functions in the code :))

PR #207, opened 2024-08-15
feat(tasks)!: Support for tasks through GraphQL, service layer and database access layer

samiralavi:
This massive PR enables end-to-end job execution and querying of status through the GraphQL interface down the database layer. I have refactored a lot of extension execution code to fit the rest of the changes for task retrieval from the Aqueduct database, both directly through our ORM and also the celery backend. Different celery flags are enabled to enable us to retrieve the `kwargs` from the task. The only data point removed from the GraphQL interface is `runtime` which showed how long the task took to execute. This can be implemented easily by enabling time tracking on the execution of extension in the service layer, but this was too much for this PR. I assume this data is not needed now and can be added later when needed to be shown on the GUI. 
A note on integration tests is that they don't run part of the CI, but the run successfully in the development environment. Again, adding CI work to run these tasks was too much for this PR. I am going to create tasks on JIRA to cover that.


aqueductcore/backend/models/task.py
```diff
@@ -0,0 +1,43 @@
...
+
+class TaskProcessExecutionResult(AQDModel):
+    """Result of process execution."""
+
+    result_code: Optional[int] = None
+    std_err: Optional[str] = None
+    std_out: Optional[str] = None
+    task_id: UUID
+    status: str
+    ended_at: Optional[datetime] = None
```
jatinriverlane, line 18: I guess we can also add `started_at` and other similar fields here, given that `ended_at` is currently present and not operational right now.

aqueductcore/backend/models/task.py
```diff
@@ -0,0 +1,43 @@
...
+    parameters: Optional[Dict] = None
+    result_code: Optional[int] = None
+    std_err: Optional[str] = None
+    std_out: Optional[str] = None
+    status: str
+    received_at: datetime
+    ended_at: Optional[datetime] = None
+
+
+class TaskCreate(TaskBase):
```
jatinriverlane, line 38: TaskCreate class inherits from TaskBase, but TaskBase has `std_err`, `std_out`, `result_code` and other similar fields. I think the fields can be redistributed among the `Create` and `Read` classes.

aqueductcore/backend/routers/graphql/types.py
```diff
@@ -181,32 +213,89 @@ class TaskStatus(Enum):
...
+class TaskData:
     """Full information about the scheduled task."""
 
     task_id: UUID = strawberry.field(description="Unique identifier of the task.")
-    experiment: Optional[ExperimentData] = strawberry.field(
-        description="Experiment the task is associated with."
+    experiment: ExperimentData = strawberry.field(
+        description="Experiment the task is associated with.", resolver=resolve_experiment
     )
-    username: Optional[str] = strawberry.field(description="User, who stated the task.")
```
jatinriverlane, line 196: There's a requirement for username of User who executed the task. How will we get this data ?

aqueductcore/backend/models/task.py
samiralavi: It is possible to refactor as you suggested, I instead set some of the fields as optional, thought it was simpler.

aqueductcore/backend/routers/graphql/inputs.py
```diff
@@ -73,3 +104,20 @@ class TasksFilterInput:
...
+
+
+@strawberry.input
+class ExecuteExtensionInput:
+    """Input for executing extension"""
+
+    experiment_uuid: UUID
+    extension: str
+    action: str
+    params: List[List[str]]
```
jatinriverlane, line 116: `List[List[str]]` is also fine, we can also use list of dictionaries.

aqueductcore/backend/routers/graphql/mutations/task_mutations.py
```diff
@@ -0,0 +1,40 @@
...
+from aqueductcore.backend.services.extensions_executor import ExtensionsExecutor
+from aqueductcore.backend.services.task_executor import revoke_task
+
+
+async def execute_extension(
+    context: ServerContext, execute_extension_input: ExecuteExtensionInput
+) -> TaskData:
+    """Extension execution mutation."""
+
+    task = await ExtensionsExecutor.execute(
```
jatinriverlane, line 18: I guess it'd be a good idea to add a `try-catch` block here so that if anything happens during execution of `execute` function, it returns valid data to the User.

jatinriverlane: Similar handling can be added to `cancel_task` function as well.

aqueductcore/backend/services/task_executor.py
```diff
@@ -84,61 +87,170 @@ def run_executable(
...
+@validate_call(config=ConfigDict(arbitrary_types_allowed=True))
+async def get_all_tasks(  # pylint: disable=too-many-arguments
+    user_info: UserInfo,
+    db_session: AsyncSession,
+    start_date: Optional[datetime] = None,
+    end_date: Optional[datetime] = None,
+    extension_name: Optional[str] = None,
+    action_name: Optional[str] = None,
+    experiment_uuid: Optional[UUID] = None,
+    order_by_creation_date: bool = True,
```
jatinriverlane, line 213: This can be `order_by_creation_time`...

aqueductcore/backend/services/task_executor.py
```diff
@@ -84,61 +87,170 @@ def run_executable(
...
+        raise AQDDBTaskNonExisting(
+            "DB query failed due to non-existing task with the specified task id."
+        )
+
+    # note: SIGINT does not lead to task abort. If you send
     # KeyboardInterupt (SIGINT), it will not stop, and the
     # exception does not propagate.
-    task.revoke(terminate=terminate, signal="SIGTERM")
-    return await update_task_info(task_id, wait=False)
+    AsyncResult(db_task.task_id).revoke(terminate=terminate, signal="SIGTERM")
```
jatinriverlane, line 166: 🤔 Maybe we can add this call to a `try-catch` block to make sure that failure of `revoke` is handled.

jatinriverlane (changes requested):
Great work @samiralavi , just a few comments from my side on this PR

aqueductcore/backend/routers/graphql/inputs.py
samiralavi: I like to do that, I kept it as it was the contract before. @safeamiiir are you happy with this change in the graphql params input type?

aqueductcore/backend/routers/graphql/mutations/task_mutations.py
samiralavi: Nice suggestion, I will go through them and try to add more exception handling

aqueductcore/backend/services/task_executor.py
samiralavi: I followed the same naming we did for experiments, so either I have to change that, or keep it coherent. I prefer to keep it as it is.

aqueductcore/backend/routers/graphql/mutations/task_mutations.py
samiralavi: GraphQL already handles exceptions and shows the error message in the returned error data, which UI shows in a pop up. It helps debug issues, instead of just hiding them behind a dumb message that an error has occurred.

aqueductcore/backend/routers/graphql/inputs.py
safeamiiir: This change makes more sense, I'm happy to do that.

aqueductcore/backend/models/task.py
jatinriverlane: Sure, fine for now. I'll create a PR mentioning few refactorings we need.

aqueductcore/backend/routers/graphql/mutations/task_mutations.py
jatinriverlane: Sure, makes sense. Thanks Amir

jatinriverlane (approved):
Thanks for addressing comments Amir

aqueductcore/backend/main.py
```diff
@@ -17,9 +18,14 @@
...
-    async with async_engine.begin() as conn:
-        await conn.run_sync(orm.Base.metadata.create_all)
+    try:
+        # initialise database with relations
+        async with async_engine.begin() as conn:
+            await conn.run_sync(orm.Base.metadata.create_all)
+    except IntegrityError as error:
+        if "already exists" not in str(error):
+            raise
+        # ignore if tables already exist
```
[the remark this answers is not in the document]
samiralavi: After we increased the number of threads from 1 to 4 when starting Aqueduct to support extensions, we forgot to fix the DB check. This is a quick solution so we don't see unnecessary error messages in the log. It didn't require its own PR.

aqueductcore/backend/routers/graphql/types.py
```diff
@@ -186,27 +218,97 @@ class KeyValuePair:
...
-    # obsolete fields:
-    return_code: int
+    experiment_uuid: strawberry.Private[UUID]
+
+
+@strawberry.type(description="Paginated list of tasks.")
+class Tasks:
+    """GraphQL node"""
+
+    tasks_data: List[TaskData] = strawberry.field(description="The list of tasks in this page")
```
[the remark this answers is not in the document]
samiralavi: We are using the same naming scheme as experiments, and extensions as tags. I say keep it the same for coherency.

safeamiiir (approved):

aqueductcore/backend/routers/graphql/inputs.py
samiralavi: For now we keep it as it is, but in future we will fix it.

PR #216, opened 2024-09-13
fix(scopes): resolving issues with lazy loading and access control

str-anger:
Making sure, that queries by ID and revoke query work both in tests and in frontend.

### DIFF
--- a/aqueductcore/backend/services/task_executor.py
+++ b/aqueductcore/backend/services/task_executor.py
@@ -146,30 +146,30 @@ async def revoke_task(
     """Cancel the task and update the status."""
 
     statement = (
-        select(orm.Task, orm.User, orm.Experiment)
-        .join(orm.User, orm.Task.created_by_user)
-        .join(orm.Experiment, orm.Task.experiment)
+        select(orm.Task)
+        .options(joinedload(orm.Task.created_by_user))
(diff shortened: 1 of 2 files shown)


aqueductcore/backend/services/task_executor.py
```diff
@@ -146,30 +146,30 @@ async def revoke_task(
...
+        .options(joinedload(orm.Task.created_by_user))
+        .options(joinedload(orm.Task.experiment))
         .filter(orm.Task.task_id == task_id)
     )
-
-    # TODO: this is not specified in ERD
-    # but it implies, as user cannot launch a task
-    # without edit permission.
-    if not user_info.can_edit_any_experiment():
-        statement = statement.filter(orm.Experiment.created_by == user_info.uuid)
```
samiralavi, line 159: I am wondering if you could use `orm.Task.experiment.created_by == user_info.uuid` here to resolve the issue

samiralavi (approved):

safeamiiir (approved):

aqueductcore/backend/services/task_executor.py
str-anger: Because this did not work with `joinedload`: it raises an exception

PR #219, opened 2024-09-23
fix(task-query): Fix non working filters

jatinriverlane:
## Changes
* FIxed the non working filters for get_all_tasks GraphQL Qhery


samiralavi (approved):

aqueductcore/backend/routers/graphql/resolvers/task_resolver.py
```diff
@@ -30,12 +30,22 @@ async def get_tasks(
             f"Maximum allowed limit for experiments is {MAX_EXPERIMENTS_PER_REQUEST}"
         )
 
+    if filters:
+        experiment = filters.experiment
+        if str(experiment.type) != "IDType.UUID":  # type: ignore
```
str-anger, line 35: I think you still allow `experiment` to be `None`, so this is why mypy is complaining on None.type. Probably:

```suggestion
    if filters is not None:
        experiment = filters.experiment
        if experiment is not None and experiment.type != IDType.UUID:
```

aqueductcore/backend/routers/graphql/resolvers/task_resolver.py
```diff
@@ -30,12 +30,22 @@ async def get_tasks(
...
+
     tasks = await get_all_tasks(
         user_info=context.user_info,
         db_session=context.db_session,
+        start_date=filters.start_date if filters else None,
+        end_date=filters.end_date if filters else None,
         extension_name=filters.extension_name if filters else None,
         action_name=filters.action_name if filters else None,
         username=filters.username if filters else None,
+        experiment_uuid=experiment.value if experiment else None,  # type: ignore
```
str-anger, line 48: I think `experiment` is out of scope, it is not defined it you did not pass the branch of `if filters:...`. Add `experiment = None` somewhere higher to add it to the scope.

aqueductcore/backend/services/task_executor.py
```diff
@@ -241,8 +239,10 @@ async def get_all_tasks(  # pylint: disable=too-many-arguments
     """Get list of all tasks."""
     statement = (
         select(orm.Task)
-        .options(joinedload(orm.Task.created_by_user))
-        .options(joinedload(orm.Task.experiment))
+        .join(orm.Task.created_by_user)
+        .join(orm.Task.experiment)
+        .options(selectinload(orm.Task.created_by_user))
+        .options(selectinload(orm.Task.experiment))
```
str-anger, line 245: I hope this does not create 2 joins =)

str-anger (changes requested):
Very minor change request, and I am happy to approve. Basically cleaning mypy

aqueductcore/backend/services/task_executor.py
jatinriverlane: No, it's creating the joins once. The only change in the final query is that previously the joins were `left outer` and now joins are `inner`.

aqueductcore/backend/routers/graphql/resolvers/task_resolver.py
jatinriverlane: Sure, I've added `experiment = None` to initialise `experiment` variable in `line 33`

str-anger (approved):

safeamiiir (approved):

PR #226, opened 2024-10-01
fix(tasks): redesign task list

safeamiiir:
Make experiment Tasks as the same level as experiment info

Screenshot 2024-10-02 at 11 48 23


str-anger (approved):

jatinriverlane (approved):
Thanks @safeamiiir

PR #227, opened 2024-10-07
fix(cancellation): TT-168 add graceful cancellation

str-anger:
Use SIGINT instead on SIGTERM and allow process to cancel gracefully

### DIFF
--- a/aqueductcore/backend/services/task_executor.py
+++ b/aqueductcore/backend/services/task_executor.py
@@ -42,7 +42,7 @@ celery_app = Celery(
 celery_app.conf.update(result_extended=True)
 extension_process = None  # pylint: disable=invalid-name
 
-def _worker_signal_handler(signo, _):
+def worker_signal_handler(signo, _):
     global extension_process  # pylint: disable=global-statement,global-variable-not-assigned
     if extension_process is not None:
         psutil_child_process = psutil.Process(extension_process.pid)
@@ -72,7 +72,7 @@ def run_executable(  # pylint: disable=unused-argument
         Tuple[int, str, str]: result code, std out, std error.
     """
     global extension_process  # pylint: disable=global-statement
-    signal.signal(signal.SIGINT, _worker_signal_handler)
+    signal.signal(signal.SIGINT, worker_signal_handler)
     extensions_dir = os.environ.get("EXTENSIONS_DIR_PATH", "")
     if not extensions_dir:
         raise FileNotFoundError("EXTENSIONS_DIR_PATH environment variable should be set.")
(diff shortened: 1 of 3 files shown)


aqueductcore/backend/services/task_executor.py
```diff
@@ -37,11 +40,20 @@
 
 
 celery_app.conf.update(result_extended=True)
+extension_process = None  # pylint: disable=invalid-name
```
str-anger, line 43: We need this variable to tell a handle which process to cancel

aqueductcore/backend/services/task_executor.py
```diff
@@ -103,15 +118,22 @@ async def _update_task_info(task_id: str, wait=False) -> TaskProcessExecutionRes
...
     if task_result is not None:
-        known_errors = (FileNotFoundError, TaskRevokedError)
-        if isinstance(task_result, known_errors):
-            err = str(task_result)
-            task_info.std_err = err
+        known_errors = (FileNotFoundError, TaskRevokedError, Exception)
+        if isinstance(task_result, KeyboardInterrupt):
+            # raised an uncaught exception
+            task_info.result_code = -signal.SIGINT
+            task_info.std_err = str(task_result)
```
safeamiiir, line 125: When this condition might happen?

aqueductcore/backend/services/task_executor.py
```diff
@@ -37,11 +40,20 @@
 
 
 celery_app.conf.update(result_extended=True)
+extension_process = None  # pylint: disable=invalid-name
+
+def _worker_signal_handler(signo, _):
```
jatinriverlane, line 45: The name of this function start with `_`. Private class methods start with `_`, but this does not belong to a class. Also, I guess we don't indend to use the `_` as the internal use indicator.

This can be renamed to remove `_` from start.

aqueductcore/backend/services/task_executor.py
str-anger: Good question. In current configuration it SHOULD NOT as we implement the worker and take care of it. But worker (`run_executable`) is not necessarily catching this type if exceptions, as this is a different application. So, if we use cancellation with sigint, but worker for some reason is not listening, to sigint signals, it will return this signal.

This happened a few times during debugging process, and propagated up to frontend, as this particular method was failing to process exception object instead of an expected tuple. This is a sanity check, and you are probably right this may be simplified to processing all exceptions uniformly.

aqueductcore/backend/services/task_executor.py
```diff
@@ -103,15 +118,22 @@ async def _update_task_info(task_id: str, wait=False) -> TaskProcessExecutionRes
...
+            task_info.std_err = str(task_result)
+        elif isinstance(task_result, known_errors):
+            task_info.std_err = str(task_result)
         elif task.ready():
-            code, out, err = task_result
-            task_info.result_code = code
-            task_info.std_out = out
-            task_info.std_err = err
+            # in case the result format is incorrect
+            if len(task_result) == 3:
```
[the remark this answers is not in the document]
str-anger: This would be a nice place to fall with an exception 😅

safeamiiir (approved):

jatinriverlane (approved):
Thanks for addressing the comments @str-anger

PR #228, opened 2024-10-07
fix(cancellation): adding failure processing

str-anger:
If extension process has a return code != 0, raise a `ChildProcessError` and then handle this at the `update_task_info` level. Treat arguments of exception as result.


safeamiiir (approved):

jatinriverlane (approved):
Great work @str-anger

PR #230, opened 2024-10-08
feat(task-uuid): Update identifier of Task from task_id to uuid

jatinriverlane:
# Changes
* Made relevant changes to update `task_id` to `uuid`
* Improved `humanize_list` utility function


str-anger (changes requested):
By itself, the PR is clean and self-descriptive.
1. I'd appreciate if you rebase on cancellation PRs to also take care of their changes.
2. I'd appreciate to see a migration script as well.

aqueductcore/backend/services/task_executor.py
```diff
@@ -219,7 +219,7 @@ async def get_task_by_uuid(
         select(orm.Task)
```
str-anger, line 219: I cannot comment to line 201, but I assume it needs to be `AsyncResult(db_task.uuid).revoke(terminate=terminate, signal="SIGINT")`

Same for line 202:
`task_info = await _update_task_info(task_id=db_task.uuid, wait=False)`

str-anger (changes requested):
Apart from comments inline, I think class `task.py/TaskProcessExecutionResult` also deserves field renaming. Can you please do this, and propagate this rename to the usage (i.e. method `_update_task_info`)

aqueductcore/backend/services/task_executor.py
jatinriverlane: Oops, this change was brought by the rebase. Fixed it, thanks.

str-anger (approved):
Conditional approve, class `TaskProcessExecutionResult` may also benefit from renaming of the field.

safeamiiir:
I found some changes which are missed in the `tests/integration/test_graphql_tasks.py` file:
like lines: `43`, `57`, `88`, ...
I guess they also need to be modified, right?

safeamiiir (approved):

PR #231, opened 2024-10-09
feat(task-cleanup): Deletion of experiment deletes all tasks related to it

jatinriverlane:
## Changes
* On deleting experiment, all tasks related to the task get deleted.
  * Removal of task rows from the Task table.
  * Removal of results, status and everything related to the tasks.

### DIFF
--- a/aqueductcore/backend/services/experiment.py
+++ b/aqueductcore/backend/services/experiment.py
@@ -8,7 +8,6 @@ from shutil import rmtree
 from typing import Callable, List, Optional, Tuple
 from uuid import UUID
 
-from celery import current_app
 from pydantic import ConfigDict, Field, validate_call
 from sqlalchemy import delete, func, or_, select
 from sqlalchemy.ext.asyncio import AsyncSession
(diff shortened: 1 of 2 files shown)


str-anger (approved):

aqueductcore/backend/services/experiment.py
```diff
@@ -522,11 +522,16 @@ async def remove_experiment(
         experiments_root_dir=str(settings.experiments_dir_path), experiment_uuid=experiment_uuid
     )
 
+    remove_all_experiment_tasks_statement = delete(orm.Task).where(
+        orm.Task.experiment_id == experiment_uuid
+    )
+    await db_session.execute(remove_all_experiment_tasks_statement)
```
safeamiiir, line 528: I'm basically seeing this line as the only actual change in this PR?

safeamiiir (approved):

aqueductcore/backend/services/experiment.py
jatinriverlane: Yes Amirreza, we are only removing the Task from `task` table. This will allow users to debug any failures in code later.

PR #232, opened 2024-10-09
fix: TT-178-change cancelation message

safeamiiir:
- change cancelation message
- use CircularProgress component instead of loading text

### DIFF
--- a/aqueductcore/frontend/src/__tests__/TaskDetailsModal.test.tsx
+++ b/aqueductcore/frontend/src/__tests__/TaskDetailsModal.test.tsx
@@ -53,7 +53,7 @@ test("clicking on cancel task button cancels task", async () => {
     await userEvent.click(cancelTaskConfirmationButton);
 
     await waitFor(() => {
-        const cancelTaskConfirmationMessage = queryByText("Task cancelled successfully");
+        const cancelTaskConfirmationMessage = queryByText("Cancelation request sent");
         expect(cancelTaskConfirmationMessage).toBeInTheDocument();
     });
 });
--- a/aqueductcore/frontend/src/components/organisms/TaskDetailsModal/index.tsx
+++ b/aqueductcore/frontend/src/components/organisms/TaskDetailsModal/index.tsx
@@ -289,7 +289,7 @@ function TaskDetailsModal({ isOpen, handleClose, taskId }: TaskDetailsModalProps
                             </List>
                         </Grid>
                         <Grid item>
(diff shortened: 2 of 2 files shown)


str-anger (approved):

aqueductcore/frontend/src/__tests__/TaskDetailsModal.test.tsx
```diff
@@ -53,7 +53,7 @@ test("clicking on cancel task button cancels task", async () => {
     await userEvent.click(cancelTaskConfirmationButton);
 
     await waitFor(() => {
-        const cancelTaskConfirmationMessage = queryByText("Task cancelled successfully");
+        const cancelTaskConfirmationMessage = queryByText("Cancelation request sent");
```
jatinriverlane, line 56: This message seems to be fine for Querra, which includes sending a cancellation request to some other API, but for the internal tasks, this message does not seem to be right.

Maybe we can keep this message for now and ask Tom for an updated message when he's back.

aqueductcore/frontend/src/components/organisms/TaskDetailsModal/index.tsx
```diff
@@ -289,7 +289,7 @@ function TaskDetailsModal({ isOpen, handleClose, taskId }: TaskDetailsModalProps
                             </List>
                         </Grid>
                         <Grid item>
-                            {isTaskCancellable ? loadingCancelTask ? <div>loading</div> : <CancelTaskButton
+                            {isTaskCancellable ? loadingCancelTask ? <CircularProgress sx={{ mr: 4 }} /> : <CancelTaskButton
```
jatinriverlane, line 292: The size of CircularProgress bar appears to be much bigger compared to the size of the cancellation button. How about decreasing the size of the circular progress bar ?

safeamiiir: i changed its size into 32px

aqueductcore/frontend/src/__tests__/TaskDetailsModal.test.tsx
safeamiiir: this is the ultimate information that we could receive in the response, how can I tell the cancelation has been successful or not? for example could you tell me a combination of returnCode or Status or etc?

jatinriverlane: Makes sense. In future, if there's any hard requirement we can come up with a further procedure of checking with server if the task is completely cancelled or not. 

It's fine for now.

PR #234, opened 2024-10-16
fix: db migration rename task id to UUID

safeamiiir:
DB on Aqueduct-core's been migrated and some more info has been added to the README file.


alembic/README
```diff
@@ -8,9 +8,9 @@ The following steps should be followed in the Python environment configured for
 
 > **NOTE:** Always assume the database is populated with existing data!
 
-1. Make sure the environment variables with the database connection details are loaded into the environment.
+1. Make sure the environment variables with the database connection details are loaded into the environment. We're using Azure, and we additionally need to make sure it's using the same environment variables to be connected to the database on Azure, as that's in the current state and does not require changes, but to make sure about that, we need to do the next step check.
```
str-anger, line 11:
```suggestion
1. Make sure the environment variables with the database connection details are loaded into the environment. If you are using Azure instance, you will additionally need to make sure it's using the same environment variables to be connected to the database on Azure, as that's in the current state and does not require changes, but to make sure about that, we need to do the next step check.
```

The fact we are using azure is our personal fact. This is not related to open-source repo.

str-anger (approved):
Conditionally ok, but text may be slightly different. Also, how this happened table name was different?

safeamiiir:
> Conditionally ok, but text may be slightly different. Also, how this happened table name was different?

It was a typo with the table name, missed because we didn't actually test it.

alembic/README
```diff
@@ -8,9 +8,9 @@ The following steps should be followed in the Python environment configured for
 
 > **NOTE:** Always assume the database is populated with existing data!
 
-1. Make sure the environment variables with the database connection details are loaded into the environment.
+1. Make sure the environment variables with the database connection details are loaded into the environment. If you are using Azure instance, you will additionally need to make sure it's using the same environment variables to be connected to the database on Azure, as that's in the current state and does not require changes, but to make sure about that, we need to do the next step check.
 
-2. Make sure the database is up to date with the head of changes. (If it is not this step tries to run the migration one by one):
+2. As an initial check, make sure the database is up to date with the head of changes. (If it is not this step, try to run the migration one by one):
```
jatinriverlane, line 13:
```suggestion
2. As an initial check, make sure the database is up to date with the head of changes. (If this step fails, try to run the migration one by one):
```

jatinriverlane (approved):
Thanks for fixing this @safeamiiir

PR #237, opened 2025-01-14
feat(user-token): Add Web UI token to environment variables

jatinriverlane:
## Changes
* Set token as environment variable

### DIFF
--- a/aqueductcore/backend/services/extensions.py
+++ b/aqueductcore/backend/services/extensions.py
@@ -125,7 +125,7 @@ class ExtensionAction(BaseModel):
 
         my_env = {key: str(val) for key, val in (extension.constants or {}).items()}
         my_env.update(params)
-        my_env["token"] = user_info.token
+        my_env["api_token"] = user_info.token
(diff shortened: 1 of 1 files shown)


str-anger (commented):
As mentioned in the other PR comment, this may probably be a neater change.

aqueductcore/backend/services/extensions.py
```diff
@@ -125,6 +125,7 @@ async def execute(
 
         my_env = {key: str(val) for key, val in (extension.constants or {}).items()}
         my_env.update(params)
+        my_env["api_token"] = user_info.token
         my_env["aqueduct_url"] = extension.aqueduct_url
         if extension.aqueduct_api_token is not None:
             my_env["API_TOKEN"] = extension.aqueduct_api_token
```
str-anger, line 131: I hope these 2 different names (lower and upper case) don't create any problem. If you can make them consistent, this may look neater.

str-anger (approved):

safeamiiir (approved):

